PR #60, opened 2025-12-04
Fix: Support Google Docs Tabs in DocsService

### LINKED ISSUE #32: Add support for Google Docs "Tabs" in DocsService
The current implementation of `DocsService.ts` (`workspace-server/src/services/DocsService.ts`) relies on the legacy/single-tab document structure. It is currently unable to interact with the new Google Docs "Tabs" feature.

When a document contains multiple tabs, the service defaults to reading from and writing to the first tab only, effectively ignoring all other content in the document.

**Affected Methods:**
* `getText` (Read operations)
* `insertText`, `appendText`, `replaceText` (Write operations)
* `create` (Creation operations)

**Analysis:**

1.  **Read Operations (`getText`):**
    The service currently accesses `res.data.body`. In multi-tab documents, content is moved to `res.data.tabs[].documentTab.body`. Accessing `res.data.body` directly only returns the content of the first tab (or legacy body) and misses all subsequent tabs.

    *Current implementation:*
    ```typescript
    const body = res.data.body; // Fails to see 'tabs' array
    ```

2.  **Write Operations (`insertText`, etc.):**
    The `batchUpdate` requests constructed by the service do not include a `segmentId`. Without a `segmentId` in the `Location` object, the Google Docs API defaults the operation to the first tab.

    *Current implementation:*
    ```typescript
    location: { index: 1 } // Missing 'segmentId' to target specific tabs
    ```

**Proposed Changes:**

1.  **Update `getText`:**
    Refactor to check for the existence of `res.data.tabs`. If present, allow retrieval of specific tabs via a new optional `tabId` parameter, or default to iterating through tabs if needed.

2.  **Update Write Methods:**
    Update `insertText`, `appendText`, and `replaceText` to accept an optional `tabId`. If provided, this `tabId` must be passed as the `segmentId` property within the `Location` object in the API request.

    *Example fix:*
    ```typescript
    insertText: {
        location: { 
            index: locationIndex,
            segmentId: tabId // New field required
        },
        text: processedText,
    }
    ```

**References:**
* Source: `https://github.com/gemini-cli-extensions/workspace/blob/main/workspace-server/src/services/DocsService.ts`
* Google Docs API: [Working with Tabs](https://developers.google.com/docs/api/how-tos/tabs)

allenhutchison:
This PR adds support for Google Docs Tabs by updating `DocsService` to use `tabId` instead of `segmentId` and ensuring `includeTabsContent: true` is used in API calls. It also updates `replaceText` to use manual `deleteContentRange` and `insertText` operations for better tab targeting and formatting support.

Fixes #32

### DIFF
--- a/workspace-server/src/__tests__/services/DocsService.test.ts
+++ b/workspace-server/src/__tests__/services/DocsService.test.ts
@@ -193,7 +193,7 @@ describe('DocsService', () => {
             };
             mockDocsAPI.documents.batchUpdate.mockResolvedValue(mockResponse);
 
-            const result = await docsService.insertText({ documentId: 'test-doc-id', text: 'Hello', tabId: 'tab-1' });
+            await docsService.insertText({ documentId: 'test-doc-id', text: 'Hello', tabId: 'tab-1' });
 
             expect(mockDocsAPI.documents.batchUpdate).toHaveBeenCalledWith({
                 documentId: 'test-doc-id',


workspace-server/src/__tests__/services/DocsService.test.ts
```diff
@@ -183,6 +183,33 @@ describe('DocsService', () => {
...
+        it('should insert text into a specific tab', async () => {
+            const mockResponse = {
+                data: {
+                    documentId: 'test-doc-id',
+                    writeControl: {},
+                },
+            };
+            mockDocsAPI.documents.batchUpdate.mockResolvedValue(mockResponse);
+
+            const result = await docsService.insertText({ documentId: 'test-doc-id', text: 'Hello', tabId: 'tab-1' });
```
heltonduarte: _result?

heltonduarte (approved):

workspace-server/src/__tests__/services/DocsService.test.ts
allenhutchison: Fixed.

PR #81, opened 2025-12-09
feat: implement drive.downloadFile tool and tests

### LINKED ISSUE #88: [FR] Drive - Extract text from generic file
## current

I understand that currently this extension can only interact with Google Docs, Google Sheets and so on.
What if I have a .txt ot .md file? I understand that this extension can't cope with it now, ot so says my GC.

## desired

Would be nice to have a way to pull all text in a file (like `getBytes()`) via API call in a new MCP function, like `Drive::extract_text(file_id)`.
I have a small script in python that does that ATM, on an other computer, which I can provided as required.

allenhutchison:
This PR introduces the `drive.downloadFile` tool, enabling users to download files from Google Drive to their local machine.

### Key Changes
- **New Tool**: `drive.downloadFile` supports downloading binary files and provides guidance for Google Workspace files (Docs, Sheets, Slides) to use their respective text extraction tools.
- **Refactoring**: 
    - Replaced the initially proposed `drive.readFile` (which returned content) with `drive.downloadFile` (which saves to disk) to better handle large files and binary data.
    - Updated `PROJECT_ROOT` detection to be more robust by looking for `gemini-extension.json`.
- **Testing**: Added comprehensive unit tests for `drive.downloadFile`, including specific error messages for Workspace files and API error handling.

This addresses feedback from the initial review of `drive.readFile`.

Fixes #88

### DIFF
--- a/workspace-server/src/__tests__/services/DriveService.test.ts
+++ b/workspace-server/src/__tests__/services/DriveService.test.ts
@@ -12,7 +12,20 @@ import { google } from 'googleapis';
 // Mock the googleapis module
 jest.mock('googleapis');
 jest.mock('../../utils/logger');
-jest.mock('node:fs');
+jest.mock('node:fs', () => {
+  const actualFs = jest.requireActual('node:fs') as any;
+  return {
+    ...actualFs,
+    promises: {
+      ...actualFs.promises,
+      mkdir: jest.fn(),
+      writeFile: jest.fn(),
+    },
+    existsSync: jest.fn(),
+    writeFileSync: jest.fn(),
+    mkdirSync: jest.fn(),
+  };
+});
 jest.mock('node:path', () => {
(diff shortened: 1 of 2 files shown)


workspace-server/src/utils/paths.ts
```diff
@@ -5,11 +5,23 @@
  */
 
 import path from 'node:path';
+import * as fs from 'node:fs';
+
+function findProjectRoot(): string {
```
heltonduarte, line 10: Didn't you have another PR for this change?

heltonduarte (approved):

workspace-server/src/utils/paths.ts
allenhutchison: Yeah, I'm not sure why this is showing as a diff as it's the same as the file in the repo at head.

PR #139, opened 2026-01-06
fix: separate release creation and asset upload in weekly preview

allenhutchison:
This change splits the weekly preview release workflow into two jobs:
1. 'create-release': Creates the release tag (runs once).
2. 'release': Builds and uploads assets (runs in parallel).

This prevents race conditions where multiple matrix jobs attempted to
create or finalize the same release simultaneously, causing
'Too many retries' errors on macOS and Windows runners.


abhipatel12 (approved):
LGTM!

PR #169, opened 2026-01-21
fix: normalise tool names to underscores for Cursor compatibility

### LINKED ISSUE #168: Bug: Tools not callable in Cursor IDE due to dot-to-underscore name conversion
## Description

MCP tools registered with dots in their names (e.g., `docs.find`, `calendar.getEvents`) fail when called from Cursor IDE.

## Steps to Reproduce

1. Configure the Google Workspace MCP server in Cursor
2. Server connects successfully and reports all tools available
3. Attempt to call any tool (e.g., `docs.find`)
4. Tool call fails with: `MCP error -32602: Tool docs_find not found`

## Root Cause

Cursor converts dots (`.`) to underscores (`_`) when invoking MCP tools. The server registers `docs.find` but Cursor calls `docs_find`, causing a mismatch.

## Proposed Fix

Normalise tool names to use underscores during registration (type-safe implementation):

```typescript
const originalRegisterTool = server.registerTool;
server.registerTool = function(...args: Parameters<typeof originalRegisterTool>) {
    const [name, ...rest] = args;
    const normalizedName = name.replace(/\./g, '_');
    return originalRegisterTool.apply(server, [normalizedName, ...rest]);
};
```

This is backwards-compatible—clients that preserve dots will still work since underscores are valid in both cases.

## Environment

- Cursor IDE (latest)
- macOS / Node v22
- Works correctly in VS Code and Kiro with identical configuration

m0nkmaster:
## Summary

Fixes #168

Cursor IDE converts dots (`.`) to underscores (`_`) when invoking MCP tools. This causes all tool calls to fail with `Tool not found` errors because the server registers tools with dots (e.g., `docs.find`) but Cursor calls them with underscores (e.g., `docs_find`). 🤷‍♂️

## Changes

- Wrap `server.registerTool` to normalise tool names (dots → underscores) at registration time
- Add documentation note about tool naming conventions

## Backwards Compatibility

This change is backwards-compatible. Clients that preserve dots (VS Code, Gemini CLI) will continue to work because underscores are valid characters in tool names for all MCP clients.

## Testing

- Verified tools work correctly in Cursor IDE after the fix
- Verified tools continue to work in VS Code
- Server reports all 51 tools registered successfully

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -62,9 +62,9 @@ npm run test -- workspace-server/src/__tests__/GmailService.test.ts
 
 To ensure code quality and formatting consistency, run the linter and tests:
 
-````bash
+```bash
 npm run test && npm run lint
-``
+```
 
 This command will run ESLint, Prettier, all tests, and other checks as defined in the project's `package.json`.
 
@@ -89,7 +89,7 @@ To separately format the code in this project by running the following command f
 
 ```bash
 npm run format
-````
+```
 
 This command uses Prettier to format the code according to the project's style guidelines.
 


m0nkmaster:
Thanks for the review feedback @gemini-code-assist! The type safety concern has been addressed - the implementation now uses `Parameters<typeof originalRegisterTool>` and `.apply()` to preserve the original type signature.

allenhutchison:
I think we'll need a different approach here. This would be a breaking change for people who have written gemini cli policies against the current tool names, as well as anyone who has written custom slash commands that rely on our MCP names (it would also break the slash commands that we ship in the extension). 

I think a better approach would be to ship an adapter that is controlled by a command line flag so that our current names are the default and the command line flag can shift the names to use an _ instead.

allenhutchison:
I've reworked this a little to add the command line flag that I mentioned. I've set the default to be _ and the flag will switch back to dot notation. The default for the gemini cli extension will continue to be dot notation.

workspace-server/src/index.ts
```diff
@@ -78,6 +78,17 @@ async function main() {
...
+    // Handle tool name normalization (dots to underscores) by default, or use dots if --use-dot-names is passed.
+    const useDotNames = process.argv.includes('--use-dot-names');
+    const separator = useDotNames ? '.' : '_';
+    
+    const originalRegisterTool = server.registerTool;
+    (server as any).registerTool = function(...args: any[]) {
+        const [name, ...rest] = args;
+        const normalizedName = name.replace(/\./g, separator);
+        return (originalRegisterTool as any).apply(server, [normalizedName, ...rest]);
+    };
```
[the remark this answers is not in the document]
allenhutchison: Good suggestion. Fixed.

workspace-server/src/__tests__/verify-naming.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+try {
+    testSeparator(false);
+    testSeparator(true);
+    console.log("Verification successful!");
+} catch (error) {
+    console.error("Verification failed:", error);
+    process.exit(1);
+}
```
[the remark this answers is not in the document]
allenhutchison: Fixed.

docs/development.md
```diff
@@ -63,7 +63,7 @@ To ensure code quality and formatting consistency, run the linter and tests:
 
 ```bash
```
abhipatel12: nit: this is not rendering properly in the rich display

workspace-server/src/__tests__/tool-normalization.test.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
abhipatel12, line 3: nit: 2026

workspace-server/src/utils/tool-normalization.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
```
abhipatel12, line 3: nit: 2026

docs/development.md
allenhutchison: Fixed.

workspace-server/src/utils/tool-normalization.ts
allenhutchison: fixed

workspace-server/src/__tests__/tool-normalization.test.ts
allenhutchison: fixed

m0nkmaster:
Thanks guys, I appreciate it! @allenhutchison  @abhipatel12

PR #210, opened 2026-02-11
Allowing command line authentication with JSON copy-and-paste

karmel:
This allows me to run the MCP server from within a command line tool and still auth-- a link is presented to the user, the user visits the link and is forwarded to a json blob which the user then pastes back at the command line. 

This code was written by Antigravity, and it worked, so I didn't look too closely :) Take what's useful, or ignore.


jpoehnelt:
also see #213 😄

allenhutchison:
#227 is going to help us to solve this in a slightly different way.

PR #222, opened 2026-02-15
Security: Fix Path Traversal in Download Tools

007vasy:
This PR fixes path traversal vulnerabilities identified in several download tools. It introduces a `resolveSafePath` utility that validates and restricts file operations to the project root directory.


allenhutchison:
Restricting file writes to the project root of the mcp server doesn't really make sense in the case of an agent using this server. Writes would need to be restricted to the root of the calling agents workspace, not the root of the MCP server workspace.

PR #225, opened 2026-02-15
Security: Redact PII in Logs

007vasy:
This PR redacts sensitive PII (emails, names, subjects) from debug logs to prevent accidental leakage.


allenhutchison:
Debug logs are off by default and when they are turned on this data is necessary for debugging.

PR #233, opened 2026-02-22
Feat sheets insert text 

### LINKED ISSUE #232: feature request: sheets.insertText
It would be nice to have a `sheets.insertText` feature to write data to a google sheet

raybell-md:
Closes https://github.com/gemini-cli-extensions/workspace/issues/232

This is a jules PR (see https://github.com/raybell-md/workspace/issues/8)

~I can strip out the changes to logger if desired~

### DIFF
--- a/workspace-server/src/__tests__/services/SheetsService.test.ts
+++ b/workspace-server/src/__tests__/services/SheetsService.test.ts
@@ -41,6 +41,7 @@ describe('SheetsService', () => {
         get: jest.fn(),
         values: {
           get: jest.fn(),
+          update: jest.fn(),
         },
       },
     };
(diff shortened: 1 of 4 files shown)


raybell-md:
To test:

1. Create a google sheet
2. Prompt

I have csv data like
```
a,b
1,2
```
Can you write that to {{ SHEET URL }}

Works albeit one cell at at time. Which is fine for this PR. We can add other function e.g. batch write in other PR's

Screenshot 2026-02-25 164523
Screenshot 2026-02-25 164622

raybell-md:
Friendly ping @allenhutchison for a review when you have time please

allenhutchison:
We can't take this PR right now because it requires new scopes in the default project. I'm working on a mechanism to allow for user-configurables scopes so that we can take features that aren't supported by the main GCP project. In those cases a user who uses these features will have to host their own GCP project to go along with it until I can get the shared GCP project updated with new scopes.

raybell-md:
Thanks for the feedback @allenhutchison . I'm not sure if helps or not but changing sheets from `spreadsheet.readonly` to `spreadsheet` brings it in line with docs which currently has write access (documents). Curious why docs write access was approved but sheets write access requires more scrutiny? I don't think you have to do anything to the underlying GCP project to expand a scope of an existing function. I was able to test fine. I understand if a new service like Keep or Tasks require a new scope it may require a change in the GCP project. Not that I recommend it but I guess you revert `documents` to `documents.readonly` to make this project read-only and allow users to expand their own scopes.

allenhutchison:
The scopes we launched with were part of the initial security review. It's not that new scopes require more scrutiny, it's that they require approval to enable in the GCP project, which requires another security review. That all requires time.

Sum1cares (approved):

PR #235, opened 2026-02-23
feat(slides): add write tools for Google Slides

### LINKED ISSUE #234: Feature: Add write tools for Google Slides
The current Slides tools [1] are all read-only (getText, find, getMetadata, getImages, getSlideThumbnail). It would be useful to also have write support.

Me and my team use Google Slides templates heavily and being able to programmatically create presentations, replace placeholder text, add slides, insert images/tables, and update formatting would save us a lot of manual work.

The tools I'm thinking of:

- `slides.create` [2] - create a new presentation
- `slides.addSlide` [3] / `deleteSlide` [4] / `duplicateSlide` [5] / `reorderSlides` [6]
- `slides.getSpeakerNotes` [7] / `updateSpeakerNotes` [7]
- `slides.replaceAllText` [8] / `insertText` [9] / `deleteText` [10]
- `slides.addShape` [11] / `addImage` [12] / `addTable` [13]
- `slides.updateTextStyle` [14] / `updateShapeProperties` [15]

All of these map to existing Google Slides API v1 `presentations.batchUpdate` requests [16]. The only scope change needed is `presentations.readonly` to `presentations` [17].

[1]: https://github.com/gemini-cli-extensions/workspace/blob/main/workspace-server/src/services/SlidesService.ts
[2]: https://developers.google.com/slides/api/reference/rest/v1/presentations/create
[3]: https://developers.google.com/slides/api/reference/rest/v1/presentations/batchUpdate#CreateSlideRequest
[4]: https://developers.google.com/slides/api/reference/rest/v1/presentations/batchUpdate#DeleteObjectRequest
[5]: https://developers.google.com/slides/api/reference/rest/v1/presentations/batchUpdate#DuplicateObjectRequest
[6]: https://developers.google.com/slides/api/reference/rest/v1/presentations/batchUpdate#UpdateSlidesPositionRequest
[7]: https://developers.google.com/slides/api/reference/rest/v1/presentations.pages#Page.FIELDS.notes_properties
[8]: https://developers.google.com/slides/api/reference/rest/v1/presentations/batchUpdate#ReplaceAllTextRequest
[9]: https://developers.google.com/slides/api/reference/rest/v1/presentations/batchUpdate#InsertTextRequest
[10]: https://developers.google.com/slides/api/reference/rest/v1/presentations/batchUpdate#DeleteTextRequest
[11]: https://developers.google.com/slides/api/reference/rest/v1/presentations/batchUpdate#CreateShapeRequest
[12]: https://developers.google.com/slides/api/reference/rest/v1/presentations/batchUpdate#CreateImageRequest
[13]: https://developers.google.com/slides/api/reference/rest/v1/presentations/batchUpdate#CreateTableRequest
[14]: https://developers.google.com/slides/api/reference/rest/v1/presentations/batchUpdate#UpdateTextStyleRequest
[15]: https://developers.google.com/slides/api/reference/rest/v1/presentations/batchUpdate#UpdateShapePropertiesRequest
[16]: https://developers.google.com/slides/api/reference/rest/v1/presentations/batchUpdate
[17]: https://github.com/gemini-cli-extensions/workspace/blob/main/workspace-server/src/index.ts#L59

stefanoamorelli:
> [!TIP]
> Better reviewed commit-by-commit.

This adds 15 write tools for Google Slides, on top of the existing 5 read-only ones [1]. Me and my team do use Google Slides templates heavily and being able to programmatically create presentations, replace placeholder text, add slides, insert images/tables, and update formatting would save us a lot of manual work.

All new tools map to existing Google Slides API v1 `presentations.batchUpdate` requests [2]. The only scope change needed is upgrading from `presentations.readonly` to `presentations` [3], which is also reflected in the GCP setup script [4].

**NB:**
`updateTextStyle` and `updateShapeProperties` accept their style/properties as a JSON string instead of `z.record(z.any())` because the MCP SDK schema serializer does not support `z.any()` and causes `tools/list` to fail silently. The JSON is parsed at runtime with a descriptive error on invalid input.

---

### Commits

- [`80f4879`](https://github.com/stefanoamorelli/workspace/commit/80f4879e531f985e962af71ef9dfcb21796dfe48) upgrade OAuth scope to full presentations access
- [`0c7a6d1`](https://github.com/stefanoamorelli/workspace/commit/0c7a6d17c45c69a05aa62e60edaa4cf169998acb) align GCP setup script scope with index.ts
- [`a1e91ca`](https://github.com/stefanoamorelli/workspace/commit/a1e91cac845bcd0c501c92f42552781924a5a785) add formatError/formatResult helpers and test mock setup
- [`f9d42ef`](https://github.com/stefanoamorelli/workspace/commit/f9d42ef9de89dc78589c5c5798dba21b14303a27) add slides.create
- [`77cc965`](https://github.com/stefanoamorelli/workspace/commit/77cc9658d616da3bd9e1d913d2b5e175e38a9f6c) add slides.addSlide
- [`3a56dd9`](https://github.com/stefanoamorelli/workspace/commit/3a56dd942e47ffd6393f07e4dcee7ca25d8339d5) add slides.deleteSlide
- [`639442c`](https://github.com/stefanoamorelli/workspace/commit/639442c48cd10a104f28fbabc58dbb5a19fb7387) add slides.duplicateSlide
- [`42361f6`](https://github.com/stefanoamorelli/workspace/commit/42361f62b9493ef804dbe07f98585870bcd07126) add slides.reorderSlides
- [`92f5eb1`](https://github.com/stefanoamorelli/workspace/commit/92f5eb1337174f66b414a88906de8a9d2badfe80) add slides.getSpeakerNotes
- [`656d145`](https://github.com/stefanoamorelli/workspace/commit/656d145954eff85a54cc39f963e38b0520a5dc7d) add slides.updateSpeakerNotes
- [`adaf091`](https://github.com/stefanoamorelli/workspace/commit/adaf09166aed18324a6949f67b1026d421fea02d) add slides.replaceAllText
- [`79d7d41`](https://github.com/stefanoamorelli/workspace/commit/79d7d41f15da06ff110249713b567491a7a8da29) add slides.insertText
- [`18d1018`](https://github.com/stefanoamorelli/workspace/commit/18d10188c70dd7b22ee6050562bbea00793e9d55) add slides.deleteText
- [`290830a`](https://github.com/stefanoamorelli/workspace/commit/290830a5638d346c15f31af50ef95d6ac5bdac99) add slides.addShape
- [`9755bf9`](https://github.com/stefanoamorelli/workspace/commit/9755bf9531be80aa04a9b4989fd72051711c5c55) add slides.addImage
- [`86eb02a`](https://github.com/stefanoamorelli/workspace/commit/86eb02a68434afddf9c87ccc597d1807d1adff79) add slides.addTable
- [`14cb5dd`](https://github.com/stefanoamorelli/workspace/commit/14cb5dd9e48e2f54c7c51cc4108742067f7454de) add slides.updateTextStyle
- [`69689c5`](https://github.com/stefanoamorelli/workspace/commit/69689c508fb8466e802f677eee596dce7516ab9f) add slides.updateShapeProperties

---

Closes #234

[1]: https://github.com/gemini-cli-extensions/workspace/blob/main/workspace-server/src/services/SlidesService.ts
[2]: https://developers.google.com/slides/api/reference/rest/v1/presentations/batchUpdate
[3]: https://github.com/gemini-cli-extensions/workspace/blob/main/workspace-server/src/index.ts#L59
[4]: https://github.com/gemini-cli-extensions/workspace/blob/main/scripts/setup-gcp.sh#L104


workspace-server/src/services/SlidesService.ts
```diff
@@ -344,6 +344,852 @@ export class SlidesService {
...
+        : '';
+
+      if (existingText.length > 0) {
+        requests.push({
+          deleteText: {
+            objectId: speakerNotesObjectId,
+            textRange: { type: 'ALL' },
+          },
+        });
+      }
```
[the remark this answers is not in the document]
stefanoamorelli: addressed :heavy_check_mark:

workspace-server/src/index.ts
```diff
@@ -431,6 +431,421 @@ async function main() {
...
+          .describe(
+            'The 0-based start index of the text range to delete.',
+          ),
+        endIndex: z
+          .number()
+          .optional()
+          .describe('The 0-based end index of the text range to delete.'),
+        type: z
+          .string()
+          .optional()
```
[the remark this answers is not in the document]
stefanoamorelli: addressed :heavy_check_mark:

workspace-server/src/index.ts
```diff
@@ -431,6 +431,421 @@ async function main() {
...
+          ),
+        endIndex: z
+          .number()
+          .optional()
+          .describe(
+            'The 0-based end index of the text range (required for FIXED_RANGE type).',
+          ),
+        type: z
+          .string()
+          .optional()
```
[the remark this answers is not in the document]
stefanoamorelli: addressed :heavy_check_mark:

workspace-server/src/index.ts
```diff
@@ -431,6 +431,421 @@ async function main() {
...
+          ),
+        layoutId: z
+          .string()
+          .optional()
+          .describe(
+            'The ID of a specific layout to use. Use slides.getMetadata to find available layouts.',
+          ),
+        predefinedLayout: z
+          .string()
+          .optional()
```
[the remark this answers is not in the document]
stefanoamorelli: addressed :heavy_check_mark:

allenhutchison:
We can't take this PR right now because it requires new scopes in the default project. I'm working on a mechanism to allow for user-configurables scopes so that we can take features that aren't supported by the main GCP project. In those cases a user who uses these features will have to host their own GCP project to go along with it until I can get the shared GCP project updated with new scopes.

stefanoamorelli:
Thanks @allenhutchison for having a look, makes sense!

When building and testing this feature I used the script available on `scripts/setup-gcp.sh` to setup my own GCP project.

Should we keep this PR open until we’d be ready to setup new scopes on the default GCP project? Happy to help rebasing if/when the time come.

allenhutchison:
Yes let's keep it open.

See #255 for discussion on this feature.

Sum1cares (approved):

Sum1cares (approved):

Sum1cares:
That's cool

Sum1cares (approved):

Sum1cares (approved):

Sum1cares (approved):

Sum1cares (approved):

Sum1cares (approved):

stefanoamorelli:
@allenhutchison now that https://github.com/gemini-cli-extensions/workspace/issues/255 is closed, should we move forward with this?

(appreciate for your feedback @Sum1cares, we'd need an approval from `CODEOWNERS` to proceed)

Sum1cares (approved):

workspace-server/src/index.ts
```diff
@@ -476,6 +476,430 @@ async function main() {
     slidesService.getSlideThumbnail,
   );
 
+  server.registerTool(
```
allenhutchison: **[blocking]** All 15 new tool registrations (this `slides.create` plus every `server.registerTool` call through line 876) bypass the local `registerTool` wrapper defined at line 172. The wrapper is what gates registration on `enabledTools` — every other tool in this file goes through it.

Worst-of-both-worlds consequence: `slides.write` is intentionally `defaultEnabled: false` in `feature-config.ts:230`, but with this bypass the 14 write tools get registered regardless. Meanwhile `SCOPES` is computed from `resolveFeatures().requiredScopes` which respects the gate, so the write scope is *not* requested — every call to a write tool will then fail with 403 at runtime. Even `WORKSPACE_FEATURE_OVERRIDES=slides.write:on` won't fix it, because the tools were already registered with the wrong scope set.

`slides.getSpeakerNotes` at line 600 has the same bug — it lands in the default-on read group so visible behavior is correct, but `slides.read:off` overrides are silently ignored.

Fix: `s/server.registerTool/registerTool/` on all 15 registrations. Might also be worth adding a test that asserts every registered tool name appears in `FEATURE_GROUPS` so this regression can't recur.

workspace-server/src/services/SlidesService.ts
```diff
@@ -280,6 +280,849 @@ export class SlidesService {
     }
   };
 
+  private formatError(method: string, error: unknown) {
+    const errorMessage =
+      error instanceof Error ? error.message : String(error);
+    logToFile(`[SlidesService] Error during ${method}: ${errorMessage}`);
+    return {
```
allenhutchison, line 374: **[blocking]** `formatError` is missing `isError: true` on the returned object. The project convention (see `DriveService.ts:38–56` and `DocsService.ts:113–121`) is that MCP error responses set `isError: true` as the discriminator. Without it, every error from every Slides tool — auth failures, 404s, 403s, JSON.parse failures, slide-not-found — gets reported to the model/client as a *successful* tool call whose content happens to contain an `error` key. The model will treat it as success and continue confidently. This is the same issue that bites the older inline catches in `getText` / `getMetadata` / `getImages` / `getSlideThumbnail`, but at least those are pre-existing.

One-line fix here propagates to all 15 new tools:
```ts
return {
  isError: true,
  content: [{ type: 'text' as const, text: JSON.stringify({ error: errorMessage }) }],
};
```

workspace-server/src/services/SlidesService.ts
```diff
@@ -280,6 +280,849 @@ export class SlidesService {
...
+      }
+
+      const response = await slides.presentations.batchUpdate({
+        presentationId: id,
+        requestBody: {
+          requests: [{ createSlide: createSlideRequest }],
+        },
+      });
+
+      const slideObjectId =
```
allenhutchison: **[blocking]** `response.data.replies?.[0]?.createSlide?.objectId ?? null` returns success with `slideObjectId: null` whenever the API reply array is empty/malformed (partial success, throttle response, schema drift). The caller/model believes the slide was created, then any follow-up tool call using the null ID 400s with a confusing error.

Same pattern in `duplicateSlide:436`, `addShape:835`, `addImage:904`, `addTable:976` — would suggest throwing instead:
```ts
const slideObjectId = response.data.replies?.[0]?.createSlide?.objectId;
if (!slideObjectId) {
  throw new Error('createSlide returned no objectId; batchUpdate reply was empty or malformed.');
}
```

workspace-server/src/services/SlidesService.ts
```diff
@@ -280,6 +280,849 @@ export class SlidesService {
...
+              replaceAllText: {
+                containsText: { text: findText, matchCase },
+                replaceText,
+              },
+            },
+          ],
+        },
+      });
+
+      const occurrencesChanged =
```
allenhutchison: **[blocking]** Same `?? 0` pattern as the slideObjectId case above — `occurrencesChanged` reports `0` both when (a) Google legitimately found zero matches and (b) the reply array is missing entirely (API change, partial failure). Users will spend hours debugging template variables that "weren't found" when the real problem was an API/permissions failure. Should distinguish missing reply (throw) from `occurrencesChanged === 0` (legitimate).

workspace-server/src/services/SlidesService.ts
```diff
@@ -280,6 +280,849 @@ export class SlidesService {
...
+            );
+            if (notesShape?.shape?.text) {
+              notesText = this.extractTextFromTextContent(
+                notesShape.shape.text,
+              ).trim();
+            }
+          }
+
+          return {
+            slideIndex: index + 1,
```
allenhutchison: **[blocking]** Returns `slideIndex: index + 1` (1-based), but every other index on the API surface — `addSlide.insertionIndex`, `reorderSlides.insertionIndex`, etc. — is 0-based. The tool description at `index.ts:603` doesn't document the difference, so an LLM caller will reasonably assume 0-based and feed this back into other tools off-by-one. Either drop the `+ 1` or call out the 1-based indexing in the description (with a strong nudge to the former for interop).

workspace-server/src/__tests__/services/SlidesService.test.ts
```diff
@@ -401,4 +403,951 @@ describe('SlidesService', () => {
       expect(response.error).toBe('API Error');
     });
   });
+
+  describe('create', () => {
```
allenhutchison, line 407: **[blocking — test gaps]** Coverage is broad (every tool has happy + error path) but several non-trivial branches in the new code are uncovered:

- `updateSpeakerNotes` — the `Speaker notes object not found` throw at `SlidesService.ts:577–581` (sibling to the already-tested `Slide not found` path)
- `updateSpeakerNotes` — the empty-notes / no-existing-text branches at 590–614 (skip delete when no `textElements`, skip insert when `notes.length === 0`, skip whole `batchUpdate` when `requests` is empty)
- `deleteText` — the `FROM_START_INDEX` branch at 750–752
- `addSlide` — the `layoutId` vs `predefinedLayout` precedence and the `objectId` parameter
- Empty/missing `replies` array on `addSlide`/`duplicateSlide`/`addShape`/`addImage`/`addTable`/`replaceAllText` — adding one test here would catch the silent-null issue I flagged above

Not a huge lift since the mocking harness is already in place. The `updateTextStyle`/`updateShapeProperties` invalid-JSON tests are exactly the right pattern — would love to see it extended.

workspace-server/src/services/SlidesService.ts
```diff
@@ -280,6 +280,849 @@ export class SlidesService {
...
+    );
+    try {
+      const id = extractDocId(presentationId) || presentationId;
+      const slides = await this.getSlidesClient();
+
+      let parsedStyle: slides_v1.Schema$TextStyle;
+      if (typeof style === 'string') {
+        try {
+          parsedStyle = JSON.parse(style);
+        } catch {
```
allenhutchison: **[important]** The `catch {}` here (and at 1089 for `updateShapeProperties`) discards the underlying parse error. The user gets "Invalid JSON for style parameter" but not "Unexpected token } at position 17". For a hand-written style string with a trailing comma or unquoted key that means iterative guessing. Could we surface the parser message?
```ts
} catch (e) {
  const detail = e instanceof Error ? e.message : String(e);
  throw new Error(`Invalid JSON for style parameter: ${detail}. Expected a JSON string like '{"bold": true}'.`);
}
```

workspace-server/src/services/SlidesService.ts
```diff
@@ -280,6 +280,849 @@ export class SlidesService {
...
+      `[SlidesService] Updating text style for object ${objectId} in presentation: ${presentationId}`,
+    );
+    try {
+      const id = extractDocId(presentationId) || presentationId;
+      const slides = await this.getSlidesClient();
+
+      let parsedStyle: slides_v1.Schema$TextStyle;
+      if (typeof style === 'string') {
+        try {
+          parsedStyle = JSON.parse(style);
```
allenhutchison: **[important]** `JSON.parse('"hello"')`, `JSON.parse('null')`, `JSON.parse('[1,2]')` all succeed and yield non-objects, which then get forwarded as `style`/`shapeProperties` to Google — the caller gets an opaque 400 instead of a clear local error. Could add an `typeof parsedStyle === 'object' && parsedStyle !== null && !Array.isArray(parsedStyle)` assertion after the parse. Same applies to `parsedProps` at line 1088.

workspace-server/src/services/SlidesService.ts
```diff
@@ -280,6 +280,849 @@ export class SlidesService {
...
+    type?: string;
+  }) => {
+    logToFile(
+      `[SlidesService] Deleting text from object ${objectId} in presentation: ${presentationId}`,
+    );
+    try {
+      const id = extractDocId(presentationId) || presentationId;
+      const slides = await this.getSlidesClient();
+
+      const textRange: slides_v1.Schema$Range = { type };
```
allenhutchison: **[important]** `type` is typed as `string` and not validated against the enum. A typo like `'all'` lowercase or `'FIXED-RANGE'` skips both `if` branches and sends `{ type: 'all' }` to Google, which produces a cryptic server-side error. Zod enforces this at the tool boundary but the service shouldn't trust callers — would add an explicit enum check here (same for `updateTextStyle` at 1029).

workspace-server/src/services/SlidesService.ts
```diff
@@ -280,6 +280,849 @@ export class SlidesService {
...
+      if (type === 'FIXED_RANGE') {
```
allenhutchison: **[important]** When `type === 'FIXED_RANGE'` and the caller forgets to pass `startIndex`/`endIndex`, the range becomes `{ type: 'FIXED_RANGE', startIndex: undefined, endIndex: undefined }` — `JSON.stringify` drops them and Google then either errors confusingly or treats it as `ALL`. Could the Zod schema make a discriminated union on `type` (`FIXED_RANGE → {start, end}`, `FROM_START_INDEX → {start}`, `ALL → {}`) so the missing-index case is unrepresentable?

workspace-server/src/index.ts
```diff
@@ -476,6 +476,430 @@ async function main() {
...
+          .describe('The ID or URL of the presentation.'),
+        objectId: z
+          .string()
+          .describe(
+            'The object ID of the shape or table cell containing the text.',
+          ),
+        style: z
+          .string()
+          .describe(
+            'A JSON string representing the text style to apply (e.g., \'{"bold": true, "fontSize": {"magnitude": 18, "unit": "PT"}}\').',
```
allenhutchison: **[important]** The `style` description gives one example but doesn't tell the caller (a) that it must conform to the Google Slides `TextStyle` schema with a link, or (b) what `fields` values are valid (`bold`, `italic`, `fontSize`, `foregroundColor`, etc., or `*` for all). LLM callers will guess field names and silently no-op when `style` and `fields` disagree. Same gap on `slides.updateShapeProperties` at line 891 — neither names `ShapeProperties` nor enumerates the valid mask values (`shapeBackgroundFill`, `outline`, `shadow`, `link`, `contentAlignment`, `autofit`). The `deleteText.type` description at line 709 is the model — it enumerates each value and its required companion params clearly.

workspace-server/src/services/SlidesService.ts
```diff
@@ -280,6 +280,849 @@ export class SlidesService {
...
+      logToFile(
+        `[SlidesService] Created presentation: ${result.presentationId}`,
+      );
+      return this.formatResult(result);
+    } catch (error) {
+      return this.formatError('slides.create', error);
+    }
+  };
+
+  public addSlide = async ({
```
allenhutchison, line 419: **[important]** Zod and the service-layer types are diverging:
- `addSlide` Zod uses `predefinedLayout: z.enum([...10 literals])`, but the service signature here widens to `predefinedLayout?: string`.
- `deleteText`/`updateTextStyle` Zod constrain `type` to the 3-literal enum but the service uses `string`.
- `updateTextStyle.style` and `updateShapeProperties.shapeProperties` are typed `string | slides_v1.Schema$TextStyle` in the service, but the Zod schema only ever accepts `string` — the object branch is dead code.

Would `z.infer<typeof schema>` work as the single source of truth? Compile-time guarantees from the Zod boundary are getting lost the moment values cross into the service.

workspace-server/src/index.ts
```diff
@@ -476,6 +476,430 @@ async function main() {
...
+          .describe('The title for the new presentation.'),
+      },
+    },
+    slidesService.create,
+  );
+
+  server.registerTool(
+    'slides.addSlide',
+    {
+      description:
```
allenhutchison, line 530: **[important]** Description says `Adds a new slide ... Optionally specify position and layout.` but doesn't mention that the tool returns the new `slideObjectId`. Callers that need to chain (e.g. `addShape` on the new slide) won't know they can read it from the response. Compare to `slides.create` which does document its return.

workspace-server/src/index.ts
```diff
@@ -476,6 +476,430 @@ async function main() {
...
+          ),
+      },
+    },
+    slidesService.updateSpeakerNotes,
+  );
+
+  server.registerTool(
+    'slides.replaceAllText',
+    {
+      description:
```
allenhutchison, line 662: **[important]** Wrapper defaults `matchCase: true`, which is opposite to Google's API default (`false`). The current wording (`case-sensitive (default: true)` lower in the schema) is correct in isolation but the description doesn't call out that omitting `matchCase` gives case-sensitive matching — a behavior surprise for anyone familiar with the underlying API.

workspace-server/src/services/SlidesService.ts
```diff
@@ -280,6 +280,849 @@ export class SlidesService {
...
+        requests.push({
+          insertText: {
+            objectId: speakerNotesObjectId,
+            insertionIndex: 0,
+            text: notes,
+          },
+        });
+      }
+
+      if (requests.length > 0) {
```
allenhutchison: **[nit]** When both `requests` end up empty (no existing notes to delete *and* `notes === ''`), this skips the `batchUpdate` and still returns the success payload at 619. Defensible as a true no-op, but there's no signal to the caller that nothing happened. Worth a `noOp: true` field on the result, or at least a debug log.

workspace-server/src/index.ts
```diff
@@ -476,6 +476,430 @@ async function main() {
...
+      inputSchema: {
+        presentationId: z
+          .string()
+          .describe('The ID or URL of the presentation.'),
+        slideObjectId: z
+          .string()
+          .describe('The object ID of the slide to add the image to.'),
+        imageUrl: z
+          .string()
+          .describe('The URL of the image to insert. Must be publicly accessible.'),
```
allenhutchison: **[nit]** "Must be publicly accessible" is the most common gotcha but Google also requires HTTPS, ≤50MB, ≤25MP, and PNG/JPEG/GIF only. Worth a one-line addition since these are common failure modes that come back as opaque API errors.

stefanoamorelli:
@allenhutchison thanks for this. I've rebased and amended to address all your feedback. Ready for your eyes.

allenhutchison:
Re-reviewed against the latest push (rebased onto current `main`) — you've addressed **everything** from the May 26 review. Verified each item:

**Blocking — all resolved ✅**
- All write tools now go through the gating `registerTool` wrapper (no direct `server.registerTool` calls)
- `formatError` sets `isError: true`
- `createSlide`/`duplicateObject`/`createShape`/`createImage`/`createTable` now throw on empty/malformed `replies` instead of returning a fake-success null `objectId`
- `replaceAllText` distinguishes a missing reply (throws) from a legitimate zero-match `occurrencesChanged`
- `getSpeakerNotes` returns 0-based `slideIndex`
- Test coverage filled in for the branches I flagged (empty-replies, speaker-notes-not-found, empty-notes skip, `FROM_START_INDEX` ±startIndex, layout precedence, invalid layout) — nicely done

**Important — key ones resolved ✅**
- Shared JSON-parse helper validates non-object/array/null with the parser detail surfaced
- `RANGE_TYPES` + discriminated range type with explicit validation throws

I pushed one small follow-up commit on top (`style(slides): apply prettier formatting to write tools`) — three files had formatting that would've failed the `format:check` CI step. I also rebased onto the latest `main` so it picks up the `ts-jest`/gaxios fix that recently landed. Hope you don't mind me handling those directly to save a round-trip.

Verified locally on the rebased branch: **575 tests pass (28 suites), ESLint clean, Prettier clean.**

Thanks for the thorough and responsive iteration on this — really nice work. Will get it merged once CI confirms green.

allenhutchison (approved):

Issue #255, opened 2026-03-03
RFC: Feature Configuration Service — Read/Write Scope-Based Feature Toggles


allenhutchison:
## Summary

We need a configuration mechanism to let users control which services and scopes the extension requests. This solves two problems:

1. **Contributor PRs that need new scopes** — PRs like #106 (Google Tasks), #237/#235 (Slides write), and #233 (Sheets write) add features requiring scopes that aren't enabled in the published GCP project. Contributors should be able to develop and test these features with their own GCP projects, while the features remain opt-in for other users.

2. **Users want scope control** — Issue #86: users who only need a few services don't want to auth for all scopes. Issue #111: users want to restrict Gmail to read-only to prevent accidental sends.

## Proposed Design: Read/Write Feature Groups

Each service is split into **read** and **write** groups. Read groups have no side effects and use readonly scopes. Write groups perform mutations and require elevated scopes.

| Service    | Group | Scopes                                                                        | Default |
| ---------- | ----- | ----------------------------------------------------------------------------- | ------- |
| `docs`     | read  | `documents`                                                                   | ✅ ON   |
| `docs`     | write | `documents`, `drive`                                                          | ✅ ON   |
| `drive`    | read  | `drive.readonly`                                                              | ✅ ON   |
| `drive`    | write | `drive`                                                                       | ✅ ON   |
| `calendar` | read  | `calendar.readonly`                                                           | ✅ ON   |
| `calendar` | write | `calendar`                                                                    | ✅ ON   |
| `chat`     | read  | `chat.spaces.readonly`, `chat.messages.readonly`, `chat.memberships.readonly` | ✅ ON   |
| `chat`     | write | `chat.spaces`, `chat.messages`, `chat.memberships`                            | ✅ ON   |
| `gmail`    | read  | `gmail.readonly`                                                              | ✅ ON   |
| `gmail`    | write | `gmail.modify`                                                                | ✅ ON   |
| `people`   | read  | `userinfo.profile`, `directory.readonly`                                      | ✅ ON   |
| `slides`   | read  | `presentations.readonly`                                                      | ✅ ON   |
| `slides`   | write | `presentations`                                                               | ❌ OFF  |
| `sheets`   | read  | `spreadsheets.readonly`                                                       | ✅ ON   |
| `sheets`   | write | `spreadsheets`                                                                | ❌ OFF  |
| `time`     | read  | _(none)_                                                                      | ✅ ON   |
[...]

PR #278, opened 2026-03-13
feat(docs): add advanced document tools

oskarcode:
## Summary
Adds four new Google Docs tools for advanced document manipulation.

### `docs.insertImage`
Inserts an image into a Google Doc at a specified position or at the end of the document.
- **Parameters**: `documentId`, `imageUrl`, `positionIndex` (optional), `tabId` (optional), `widthPt` (optional), `heightPt` (optional)

### `docs.insertTable`
Inserts a table into a Google Doc at a specified position or at the end of the document.
- **Parameters**: `documentId`, `rows`, `columns`, `tabId` (optional), `positionIndex` (optional)

### `docs.createHeaderFooter`
Creates a header or footer in a Google Doc with optional initial text.
- **Parameters**: `documentId`, `type` ("header" or "footer"), `text` (optional)

### `docs.addComment`
Adds a comment to a Google Doc's comment thread.
- **Parameters**: `documentId`, `content`

## Changes
- Added `insertImage`, `insertTable`, `createHeaderFooter`, `addComment` methods to `DocsService.ts`
- Added `getDriveClient` helper method for Drive API access (needed by `addComment`)
- Registered all four new tools in `index.ts`

## Testing
Tested locally with Google Docs documents.

### DIFF
--- a/workspace-server/src/services/DocsService.ts
+++ b/workspace-server/src/services/DocsService.ts
@@ -696,7 +696,7 @@ export class DocsService {
           content = tabs[0].documentTab?.body?.content;
         }
         const lastElement = content?.[content.length - 1];
-        const endIndex = lastElement?.endIndex || 2;
+        const endIndex = lastElement?.endIndex || 1;
         insertIndex = Math.max(1, endIndex - 1);
       }
 
@@ -711,14 +711,14 @@ export class DocsService {
       };
 
       if (widthPt || heightPt) {
-        imageRequest.insertInlineImage!.objectSize = {
-          width: widthPt
-            ? { magnitude: widthPt, unit: 'PT' }
-            : { magnitude: 300, unit: 'PT' },
-          height: heightPt
-            ? { magnitude: heightPt, unit: 'PT' }
-            : { magnitude: 200, unit: 'PT' },
-        };
+        const objectSize: docs_v1.Schema$Size = {};
+        if (widthPt) {
+          objectSize.width = { magnitude: widthPt, unit: 'PT' };
+        }
+        if (heightPt) {
+          objectSize.height = { magnitude: heightPt, unit: 'PT' };
+        }
+        imageRequest.insertInlineImage!.objectSize = objectSize;
       }
 
       const update = await docs.documents.batchUpdate({
@@ -793,7 +793,7 @@ export class DocsService {
           content = tabs[0].documentTab?.body?.content;
         }
         const lastElement = content?.[content.length - 1];
-        const endIndex = lastElement?.endIndex || 2;
+        const endIndex = lastElement?.endIndex || 1;
         insertIndex = Math.max(1, endIndex - 1);
       }
 
--- a/workspace-server/src/services/DocsService.ts
+++ b/workspace-server/src/services/DocsService.ts
@@ -114,8 +114,9 @@ export class DocsService {
 
   /**
    * Calculates the appropriate insertion index for new content in a document.
-   * If positionIndex is provided, uses that. Otherwise, finds the end of the document content
-   * and calculates an appropriate insertion point.
+   * If positionIndex is provided, uses that. Otherwise, finds the end of the
+   * target tab's content (or first tab when tabId is not provided) and
+   * calculates an appropriate insertion point.
    */
   private async _calculateInsertionIndex(
     documentId: string,
@@ -132,10 +133,10 @@ export class DocsService {
       fields: 'tabs',
       includeTabsContent: true,
     });
-    
-    const tabs = res.data.tabs || [];
+
+    const tabs = this._flattenTabs(res.data.tabs || []);
     let content: docs_v1.Schema$StructuralElement[] | undefined;
-    
+
     if (tabId) {
       const tab = tabs.find((t) => t.tabProperties?.tabId === tabId);
       if (!tab) {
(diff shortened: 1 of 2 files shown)


workspace-server/src/services/DocsService.ts
```diff
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-import { google, docs_v1 } from 'googleapis';
+import { google, docs_v1, drive_v3 } from 'googleapis';
```
allenhutchison: See #255 we are keeping a strict seperation of concerns for oauth scopes. All of our commenting features (which require the drive scope and api) are in the drive.* namespace. Please move commenting features there.

allenhutchison (changes requested):

workspace-server/src/services/DocsService.ts
oskarcode: Thanks for the feedback. 
- Removed Drive/comment-specific pieces from DocsService

allenhutchison (commented):
Thanks for this PR! These are great additions to the Docs toolset. I did a review and have some feedback — mostly small things, with a couple of items worth addressing before merge.

## Must Fix

### 1. `_calculateInsertionIndex` should flatten nested tabs

The new helper searches `res.data.tabs` directly without calling `this._flattenTabs()`, but every other tab-aware method in the class (`writeText`, `getText`, `replaceText`) flattens the tab tree first. This means `insertImage` and `insertTable` will throw "Tab not found" for any nested/child tab.

```diff
-    const tabs = res.data.tabs || [];
+    const tabs = this._flattenTabs(res.data.tabs || []);
```

### 2. `createHeaderFooter` silently drops text when `segmentId` is missing

If the API doesn't return a `headerId`/`footerId` for some reason, the `if (text && segmentId)` check silently skips text insertion and reports success. The user asked for a header *with text* but gets one without — and no indication anything went wrong.

Could we surface this as an error instead?

```typescript
if (text && !segmentId) {
  throw new Error(
    `Created ${type} but could not retrieve its ID from the API response. The provided text was not inserted.`,
  );
}
```

## Should Fix

### 3. Consider adding `isError: true` to error responses

The existing `getSuggestions` method includes `isError: true` in its error return, which is the MCP-correct way to signal tool failures to clients. The new methods omit it. Worth adding for consistency (or removing from `getSuggestions` if the project intentionally omits it — but including it everywhere is the better practice).

### 4. Aspect ratio comment may be misleading

The comment at line 725 says "Google Docs API will automatically preserve aspect ratio" when only one dimension is provided, and the log messages repeat this claim. I don't believe this is a documented API guarantee — the behavior when only one dimension is specified isn't well-defined. Maybe soften to something like:

```typescript
// Only set explicit dimensions if provided. If only one dimension is given,
// the API behavior for the missing dimension is not well-documented.
```

### 5. `imageUrl` could use `.url()` validation

Small one — `z.string()` for `imageUrl` could be `z.string().url()` to catch invalid URLs at the schema level instead of producing a confusing API error downstream.

## Nits (take or leave)

- `rows`/`columns` schema could benefit from a `.max()` to prevent opaque API failures with huge tables
- `widthPt || heightPt` on line 727 treats `0` as "not provided" — `!== undefined` would be more precise
- The `_calculateInsertionIndex` JSDoc doesn't mention the `tabId` parameter
- Line 150 uses `|| 1` where `?? 1` would be more semantically correct (handles `endIndex: 0`)

Overall this is solid work — the `_calculateInsertionIndex` extraction is a nice cleanup, and the `createHeaderFooter` approach with `endOfSegmentLocation` is well done. Looking forward to getting these tools in! 🎉

allenhutchison:
Hi @oskarcode! Great work on this. I noticed the CI is currently failing due to Prettier formatting checks. You can fix this by running `npm run format:fix` in the project root.

I also have a few small suggestions based on the remaining review items to get this fully polished:
1. **Tool Consistency**: Add `isError: true` to the `catch` blocks in `writeText`, `formatText`, `getText`, and `replaceText` (consistency with your new methods).
2. **Input Validation**: Add `.max(100)` to `rows` and `columns` for `docs.insertTable` in `index.ts`.
3. **Documentation**: Update the `_calculateInsertionIndex` JSDoc to include the `tabId` parameter.

Once these are updated and the CI is green, we can move forward with merging! 🎉

bdoyle0182:
would love to see these tools get released. @oskarcode do you have bandwidth to handle the feedback from Allen?

oskarcode:
> would love to see these tools get released. @oskarcode do you have bandwidth to handle the feedback from Allen?

sorry, just saw this. I will try to do this few days.

bdoyle0182:
@oskarcode are you still going to pick this up anytime soon? it would be good to know if you don't have any bandwidth

allenhutchison:
Hi @oskarcode — thanks for sticking with this one. You addressed all the substantive review feedback: the nested-tab flattening in `_calculateInsertionIndex`, the header/footer error-on-missing-segmentId, the OAuth scope separation (moving the commenting feature out of `DocsService`), plus the `isError` and `imageUrl.url()` follow-ups. The `insertImage` / `insertTable` / `createHeaderFooter` implementations themselves are in good shape.

I'm going to close this as stale, since it's been a couple of months and the branch has drifted ~57 commits behind `main`. The remaining work to get it merge-ready is small but does need your hand, since the branch is on your fork and isn't set up for maintainer edits:

1. **Rebase onto `main`.** The feature code (`DocsService.ts`) merges cleanly; the only conflict is in `index.ts`, where `main` switched tool registration from `server.registerTool(...)` to a feature-gated `registerTool(...)` wrapper (same signature — just drop the `server.` prefix on the three new registrations).
2. **Register the new tools in the feature config.** `main` now has a feature-group system (`src/features/feature-config.ts`); the three tools should be added to the `docs` write group with the appropriate scope.
3. **Run `npm run format:fix`** to clear the Prettier CI failure.

**Please feel free to reopen** (or open a fresh PR from an updated branch) whenever you'd like to finish it — the hard part is done, and I'm happy to re-review and get it merged. Thanks again for the contribution!

PR #279, opened 2026-03-13
feat(calendar): add calendar management tools

oskarcode:
## Summary
Adds three new Google Calendar tools for advanced calendar management.

### `calendar.createCalendar`
Creates a new calendar.
- **Parameters**: `summary`, `description` (optional), `timeZone` (optional)

### `calendar.createRecurringEvent`
Creates a recurring event in a calendar with RRULE support.
- **Parameters**: `calendarId` (optional), `summary`, `description` (optional), `start`, `end`, `attendees` (optional), `recurrence`, `reminders` (optional)

### `calendar.setEventReminders`
Sets custom reminders for an existing calendar event or resets to default reminders.
- **Parameters**: `eventId`, `calendarId` (optional), `useDefault` (optional), `overrides` (optional)

## Changes
- Added `createCalendar`, `createRecurringEvent`, `setEventReminders` methods to `CalendarService.ts`
- Added corresponding interfaces for input types
- Registered all three new tools in `index.ts`

## Testing
Tested locally with Google Calendar.

### DIFF
--- a/workspace-server/src/index.ts
+++ b/workspace-server/src/index.ts
@@ -639,113 +639,6 @@ async function main() {
   );
 
   server.registerTool(
-    'calendar.createCalendar',
-    {
-      description: 'Creates a new calendar.',
-      inputSchema: {
-        summary: z.string().describe('Calendar name/title.'),
-        description: z
-          .string()
-          .optional()
-          .describe('Optional calendar description.'),
-        timeZone: z
-          .string()
-          .optional()
-          .describe('Optional IANA timezone (e.g., America/New_York).'),
-      },
-    },
-    calendarService.createCalendar,
-  );
-
-  server.registerTool(
-    'calendar.createRecurringEvent',
-    {
-      description: 'Creates a recurring event in a calendar.',
-      inputSchema: {
-        calendarId: z
-          .string()
-          .optional()
-          .describe(
-            'The ID of the calendar to create the recurring event in. Defaults to primary.',
-          ),
-        summary: z.string().describe('The summary or title of the event.'),
-        description: z
-          .string()
-          .optional()
-          .describe('The description of the event.'),
-        start: z.object({
-          dateTime: z
-            .string()
-            .describe(
-              'The start time in strict ISO 8601 format with seconds and timezone.',
-            ),
-        }),
-        end: z.object({
-          dateTime: z
-            .string()
-            .describe(
-              'The end time in strict ISO 8601 format with seconds and timezone.',
-            ),
-        }),
-        attendees: z
-          .array(z.string())
-          .optional()
-          .describe('The email addresses of the attendees.'),
-        recurrence: z
-          .array(z.string())
-          .min(1)
-          .describe(
-            'Recurrence rules (e.g., ["RRULE:FREQ=WEEKLY;BYDAY=MO;COUNT=10"]).',
-          ),
-        reminders: z
-          .object({
-            useDefault: z.boolean().optional(),
-            overrides: z
-              .array(
-                z.object({
-                  method: z.enum(['email', 'popup']),
-                  minutes: z.number().int().min(0),
-                }),
-              )
-              .optional(),
-          })
-          .optional()
-          .describe('Optional reminder configuration for the recurring event.'),
-      },
-    },
-    calendarService.createRecurringEvent,
-  );
-
-  server.registerTool(
-    'calendar.setEventReminders',
-    {
-      description:
-        'Sets custom reminders for an existing calendar event (or resets to default reminders).',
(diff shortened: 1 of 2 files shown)


allenhutchison (commented):
Thanks for this PR! The new calendar tools are useful additions, and the `getErrorMessage`/`createApiErrorResponse` refactoring is a nice cleanup. I have some feedback below — a couple of things to address before merge and one design suggestion I'd love to discuss.

## Design Suggestion: Consolidate `createRecurringEvent` into `createEvent`

`createRecurringEvent` is essentially `createEvent` with `recurrence` and `reminders` added. The validation, `sendUpdates` defaulting, event construction, and `events.insert` call are all duplicated. This means:

- If a bug is fixed or feature added to `createEvent`, `createRecurringEvent` won't benefit
- `createRecurringEvent` already lacks `addGoogleMeet` and `attachments` support that `createEvent` has
- The duplication will grow as more features are added

**Suggestion:** Add `recurrence` (optional `string[]`) and `reminders` (optional) to `CreateEventInput` and handle them in `createEvent`. You can still register a separate `calendar.createRecurringEvent` tool name in `index.ts` that maps to the same service method but makes `recurrence` required in its schema. This way you get two distinct tools for LLM consumers without duplicating the implementation.

## Must Fix

### 1. `getPrimaryCalendarId()` called outside try-catch in both new methods

In `createRecurringEvent` (line 371), the call sits between the validation try-catch and the API try-catch. In `setEventReminders` (line 397), it's before the try-catch entirely. If this throws (auth failure, network error, quota), the error propagates unhandled instead of returning a structured error response.

```typescript
// Move inside the API try-catch:
try {
  const finalCalendarId = calendarId || (await this.getPrimaryCalendarId());
  // ... rest of method
} catch (error) {
  return this.createApiErrorResponse(error, 'calendar.createRecurringEvent');
}
```

(This is a pre-existing pattern in other methods too, but no need to replicate it in new code.)

### 2. `setEventReminders` silently ignores custom overrides

The destructuring `const { useDefault = true, overrides } = input` means if a user provides `overrides` but doesn't explicitly set `useDefault: false`, the API receives `{ useDefault: true, overrides: [...] }` and silently discards the custom reminders. The tool returns success but nothing changes.

```typescript
// Auto-correct when overrides are provided:
let useDefault = input.useDefault ?? true;
if (overrides?.length && useDefault) {
  logToFile('overrides provided with useDefault=true; setting useDefault to false');
  useDefault = false;
}
```

## Should Fix

### 3. Indentation mismatch in index.ts

The new tool registrations use 3-space indentation while the rest of the file uses 4-space. This also reformats the closing brace of the preceding `deleteEvent` registration and the opening of `chat.listSpaces`.

### 4. `sendUpdates` missing from tool schemas

The PR adds `sendUpdates` to `UpdateEventInput` and the `updateEvent` service logic, but the `calendar.updateEvent` schema in index.ts doesn't include it — so MCP clients can never pass it. Same for `createRecurringEvent`'s schema. The auto-default will always apply, which may not match user intent.

Add to both schemas (matching the existing `calendar.createEvent` pattern):
```typescript
sendUpdates: z
  .enum(['all', 'externalOnly', 'none'])
  .optional()
  .describe('Whether to send notifications. Defaults to "all" if attendees provided, otherwise "none".'),
```

## Nits (take or leave)

- `createApiErrorResponse` could include `isError: true` in the return — this is the MCP-correct way to signal failures (DriveService does this)
- The `sendUpdates` log `"Sending updates: none"` reads like updates *are* being sent. Also the `if (finalSendUpdates)` guard is always true after defaulting since `'none'` is truthy
- The recurrence parameter description could include a couple more RRULE examples and mention RFC 5545 — helps LLM consumers construct less common patterns
- `useDefault` Zod schema could use `.default(true)` so the JSON Schema itself reflects the default, not just the description string

Nice work overall — looking forward to the next iteration!

Sum1cares (approved):

allenhutchison:
Thanks again for this contribution, @oskarcode — the calendar tools were a genuinely useful idea and the `getErrorMessage`/`createApiErrorResponse` cleanup was nice work.

I'm closing this out as stale since there's been no activity since March and the [detailed review feedback](https://github.com/gemini-cli-extensions/workspace/pull/279#pullrequestreview) from March 19 went unaddressed — it now also has merge conflicts with `main`. No reflection on the work itself; we just don't want to leave it lingering.

If you'd like to pick it back up, please feel free to **reopen** or open a fresh PR — rebased on the latest `main`, with the must-fix items (the `getPrimaryCalendarId()` error handling and the `setEventReminders` override discard) and a few unit tests. I'd be happy to review. Thanks!

Issue #177, opened 2026-01-26
`calendar.createEvent` tool enforces `dateTime` requirement, preventing all-day event creation


AlpineMutts:
The `calendar.createEvent` tool [currently requires](https://github.com/gemini-cli-extensions/workspace/blob/5908babc2b9f84be6681caf87345658889eea1fa/workspace-server/src/services/CalendarService.ts#L17) the `dateTime` property in the `start` and `end` parameters. According to the Google Calendar API documentation, all-day events are created by providing a `date` string (formatted as `YYYY-MM-DD`) and omitting the `dateTime` field.

Source: https://developers.google.com/workspace/calendar/api/guides/create-events states:
> Specify timed events using the start.dateTime and end.dateTime fields. For all-day events, use start.date and end.date instead.

When attempting to create an all-day event by providing `date` but not `dateTime`, the tool returns a validation error: `params/start must have required property 'dateTime'`.

## Steps to Reproduce
1. Attempt to create an event using `calendar.createEvent`.
2. Provide `start: { date: "2026-01-26" }` and `end: { date: "2026-01-27" }`.
3. The tool fails with a schema validation error: `params/start must have required property 'dateTime'`.

## Expected Behavior
The tool should allow the `date` property as an alternative to `dateTime` (making both optional but requiring at least one), enabling the creation and modification of all-day events as supported by the underlying Google Calendar API.

## Impact
Users cannot use the agent to manage all-day events. This forces the agent to use timed slots (e.g., 09:00 - 17:00) as a workaround, which clutters the user's schedule grid.

PR #290, opened 2026-03-24
feat(calendar): introduce eventType (Out of Office, Focus Time, Working Location) in Calendar Service

galuszkak:
Implements Google Calendar Status Events by adding three new MCP tools - `calendar.createFocusTime`, `calendar.createOutOfOffice`, and `calendar.setWorkingLocation` - that allow users to create status events with their respective properties such as auto-decline mode, decline messages, chat status, and working location details. The `calendar.listEvents` tool is extended with an eventTypes filter parameter so users can query specific event types like `focusTime`, `outOfOffice`, `workingLocation`, birthday, or fromGmail, and its response fields now include eventType, focusTimeProperties, outOfOfficeProperties, and workingLocationProperties. This change also fixes an issue where status events without a summary were silently dropped from list results, since these event types may not always have a summary set.

### DIFF
--- a/skills/google-calendar/SKILL.md
+++ b/skills/google-calendar/SKILL.md
@@ -379,13 +379,13 @@ Users may have multiple calendars (personal, work, shared team calendars).
 
 ## Tool Quick Reference
 
-| Tool                      | Action                            | Key Parameters                                                                                  |
-| :------------------------ | :-------------------------------- | :---------------------------------------------------------------------------------------------- |
-| `calendar.list`           | List all calendars                | _(none)_                                                                                        |
-| `calendar.listEvents`     | List events (filterable by type)  | `calendarId`, `timeMin`, `timeMax`, `eventTypes`                                                |
-| `calendar.getEvent`       | Get event details                 | `eventId`, `calendarId`                                                                         |
-| `calendar.createEvent`    | Create event (all types)          | `calendarId`, `summary`, `start`, `end`, `eventType`, `addGoogleMeet`, `attachments`            |
-| `calendar.updateEvent`    | Modify an existing event          | `eventId`, `summary`, `start`, `end`, `attendees`, `addGoogleMeet`, `attachments`               |
-| `calendar.deleteEvent`    | Delete an event                   | `eventId`, `calendarId`                                                                         |
-| `calendar.respondToEvent` | Accept/decline an invite          | `eventId`, `responseStatus`                                                                     |
-| `calendar.findFreeTime`   | Find available meeting time       | `attendees`, `timeMin`, `timeMax`, `duration`                                                   |
+| Tool                      | Action                           | Key Parameters                                                                       |
+| :------------------------ | :------------------------------- | :----------------------------------------------------------------------------------- |
+| `calendar.list`           | List all calendars               | _(none)_                                                                             |
+| `calendar.listEvents`     | List events (filterable by type) | `calendarId`, `timeMin`, `timeMax`, `eventTypes`                                     |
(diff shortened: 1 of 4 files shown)


allenhutchison:
Hey @galuszkak — the `listEvents` extensions and the summary-filter bug fix look great, those should stay as-is.

However, I'd like to rethink the approach of adding three separate tools (`calendar.createFocusTime`, `calendar.createOutOfOffice`, `calendar.setWorkingLocation`). Instead, let's extend the existing `calendar.createEvent` with an `eventType` enum and the relevant type-specific optional properties.

**Why:**
- The Google Calendar API itself uses a single `events.insert` endpoint with an `eventType` field — our tool surface should mirror that rather than fan out into per-type tools.
- Three new tools means tool sprawl for LLM clients that need to choose between four creation tools, and every new event type Google adds would mean yet another tool.
- The three new service methods duplicate ~200 lines of logic (calendar ID resolution, datetime validation, `events.insert`, error/success response construction) that `createEvent` already handles.

**What I'd like to see instead:**

Add optional fields to `calendar.createEvent`'s schema:
- `eventType?: 'default' | 'focusTime' | 'outOfOffice' | 'workingLocation'`
- `focusTimeProperties?: { chatStatus, autoDeclineMode, declineMessage }`
- `outOfOfficeProperties?: { autoDeclineMode, declineMessage }`
- `workingLocationProperties?: { type, homeOffice, officeLocation, customLocation }`

Then have `createEvent` set `transparency` and `visibility` automatically based on the `eventType` (e.g., `workingLocation` → `transparency: 'transparent', visibility: 'public'`; `focusTime`/`outOfOffice` → `transparency: 'opaque'`).

**Additionally**, this is a good opportunity to fix #177 in the same PR — `createEvent`'s `start`/`end` schema currently requires `dateTime`, which prevents all-day event creation. Your `setWorkingLocation` already handles both `date` and `dateTime` — we need that same flexibility in `createEvent` itself. Making `start` and `end` accept either `date` or `dateTime` (both optional, at least one required) would unblock all-day events for all event types, including OOO and working location.

**To summarize the ask:**
1. Remove the three new tools and fold their functionality into `createEvent`
2. Update `createEvent`'s `start`/`end` to support both `date` and `dateTime` (fixes #177)
3. Keep the `listEvents` changes (eventTypes filter, expanded fields, summary-filter fix)

galuszkak:
I actually did it on purpose - given it's different form on UI of Google Calendar - it made sense to me to have this in MCP as 3 different tools - but if you are saying that this should be one tool - I have no problem changing that!

Let me get back today / tomorrow on this.

galuszkak:
@allenhutchison updated - I've also updated the skill itself - hope this is OK.

allenhutchison:
Thanks for reworking this @galuszkak — the consolidation into `createEvent` looks great. A few small things before this can land:

**Minor nits:**
- Stray blank line added between `deleteEvent` and `chat.listSpaces` tool registrations in `index.ts`
- `declineMessage: undefined` is still explicitly sent in the request body for focusTime/outOfOffice when not provided — consider conditionally including it
- `eventTypes` on `ListEventsInput` is typed as `string[]` but the Zod schema uses an enum — would be good to use the union type in the interface for consistency

**CI:** Make sure to run the formatting and lint checks locally before pushing — CI will fail otherwise:

```bash
npm run format:check   # check prettier formatting
npm run format:fix     # auto-fix formatting issues
npm run lint           # check for lint errors
npm run lint:fix       # auto-fix lint issues
```

galuszkak:
@allenhutchison - I think this is ready for review - apologise for skipping local linting.

PR #308, opened 2026-03-30
fix(calendar): tighten createEvent validation for event types

allenhutchison:
## Summary

Follow-up to #290 — tightens validation in `createEvent` to catch edge cases that could produce confusing API behavior:

- **Reject ambiguous start/end**: If both `dateTime` and `date` are provided, reject instead of letting the Google API pick one
- **Require summary for regular events**: `summary` was made optional in #290 to support status event defaults, but regular events (no `eventType` or `eventType: "default"`) should still require it
- **Validate working location sub-properties**: `officeLocation` details are now required when `type` is `"officeLocation"`, and `customLocation` details when `type` is `"customLocation"`
- **Fix updateEvent for all-day events**: `updateEvent` was unconditionally validating `start.dateTime` even for date-only updates

## Test plan

- [x] 7 new test cases covering all validation scenarios
- [x] All 74 calendar service tests pass
- [x] Lint and format checks pass

### DIFF
--- a/workspace-server/src/services/CalendarService.ts
+++ b/workspace-server/src/services/CalendarService.ts
@@ -672,6 +672,34 @@ export class CalendarService {
       attachments,
     } = input;
 
+    // Validate start/end: if provided, exactly one of dateTime or date
+    if (start) {
+      if ((start.dateTime && start.date) || (!start.dateTime && !start.date)) {
+        return this.createValidationErrorResponse(
+          new z.ZodError([
+            {
+              code: 'custom',
+              message: 'start must have exactly one of "dateTime" or "date"',
+              path: ['start'],
+            },
+          ]),
+        );
+      }
+    }
+    if (end) {
+      if ((end.dateTime && end.date) || (!end.dateTime && !end.date)) {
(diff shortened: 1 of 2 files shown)


workspace-server/src/services/CalendarService.ts
```diff
@@ -260,15 +260,42 @@ export class CalendarService {
     const summary =
       input.summary ?? (eventType ? summaryDefaults[eventType] : undefined);
 
-    // Validate start/end: at least one of dateTime or date must be provided
-    if ((!start.dateTime && !start.date) || (!end.dateTime && !end.date)) {
+    // Validate start: exactly one of dateTime or date must be provided
```
abhipatel12, line 263: The validations here make sense. At some point it may make sense to have a validation layer to separate all this logic to make it easier to read.

abhipatel12 (approved):
LGTM

workspace-server/src/services/CalendarService.ts
allenhutchison: Good point filed #313 to cover this work.

Sum1cares (approved):

PR #319, opened 2026-04-02
fix(calendar): preserve eventType when updating Focus Time, OOO, and Working Location events

RussellZager:
## Summary

`calendar.updateEvent` uses `events.update` (PUT), which replaces the entire event resource. This causes two bugs:

1. **Focus Time / OOO / Working Location events fail** with `"Event type cannot be changed"` — omitting `eventType` from the PUT body is interpreted as changing it to `default`
2. **Data loss on partial updates** — fields not included in the request (summary, description, reminders, visibility, colorId, conferenceData, etc.) get silently wiped

## Fix

Switch from `events.update` (PUT) to `events.patch` (PATCH). PATCH only modifies the fields present in the request body, preserving everything else. This matches the existing code comment (`// patch semantics`) and is the correct HTTP method for partial updates.

**1 file changed, 9 insertions, 4 deletions.**

## Repro (bug 1)

1. Create a Focus Time event in Google Calendar
2. Update it via MCP (e.g., change start/end time)
3. `{"error": "Event type cannot be changed."}`

## Repro (bug 2)

1. Create an event with a title, description, and custom reminders
2. Update only the start/end time via MCP
3. Title, description, and reminders are silently deleted

## Ref

- [events.patch docs](https://developers.google.com/calendar/api/v3/reference/events/patch)
- [eventType is immutable after creation](https://developers.google.com/calendar/api/v3/reference/events#eventType)

## Test plan

- [x] Focus Time event: updated time, title and eventType preserved
- [x] Regular event: updated time, no field loss
- [x] Build passes (`node esbuild.config.js`)

### DIFF
--- a/workspace-server/src/__tests__/services/CalendarService.test.ts
+++ b/workspace-server/src/__tests__/services/CalendarService.test.ts
@@ -857,7 +857,7 @@ describe('CalendarService', () => {
         attendees: [{ email: 'new@example.com' }],
       };
 
-      mockCalendarAPI.events.update.mockResolvedValue({ data: updatedEvent });
+      mockCalendarAPI.events.patch.mockResolvedValue({ data: updatedEvent });
 
       const result = await calendarService.updateEvent({
         eventId: 'event123',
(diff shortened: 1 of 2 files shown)


galuszkak:
I did address also this in #316

RussellZager:
Thanks @galuszkak — just saw #316 covers this as part of the larger refactor. Closing as duplicate. The core fix is the same: `events.update` (PUT) → `events.patch` (PATCH) to preserve `eventType` and avoid silent field loss on partial updates.

PR #325, opened 2026-04-07
fix(cloud_function): return full base64 state for v0.0.9+ compatibility

bs-nubank:
## Problem

Starting with **workspace-server v0.0.9**, the OAuth callback state validation was updated to decode the full base64 JSON payload returned by the cloud function, then extract the `csrf` field from it:

```javascript
// v0.0.9 — extension/server/index.js
const i = searchParams.get('state');
const A = JSON.parse(Buffer.from(i, 'base64').toString('utf8'));
const g = A?.csrf ?? null;
if (!i || g !== localCsrfToken) {
  res.end('State mismatch. Possible CSRF attack.');
}
```

However, the cloud function still returns **only the raw `csrf` hex string** in the `state` parameter:

```javascript
// cloud_function/index.js (before this fix)
if (payload.csrf) {
  finalUrl.searchParams.append('state', payload.csrf); // e.g. "daf5b690a71e..."
}
```

When the extension receives this raw hex value and tries to `Buffer.from(hex, 'base64')` → `JSON.parse`, it fails silently, sets `csrf` to `null`, and the validation **always** returns:

> `State mismatch. Possible CSRF attack.`

This makes every browser-based auth attempt on v0.0.9 fail. Users are stuck in a broken auth loop — the only workaround is the headless login (`node index.js login`).

## Root cause

The extension code was updated in v0.0.9 to a new validation protocol (decode the full state, extract csrf) but the cloud function was not updated to match — it still uses the old protocol (return only the raw csrf hex).

### 1. `fix(cloud_function)`: Return original `state` parameter unchanged

```javascript
// After this fix
if (state) {
  finalUrl.searchParams.append('state', state); // full base64 JSON, as sent originally
}
```

### 2. `fix(auth)`: Support both state formats in AuthManager for backward compatibility

> ⚠️ **Backward compatibility issue discovered after the initial commit.**

The cloud function fix changes the `state` value from raw hex to full base64 JSON. `workspace-server ≤v0.0.7` compares the returned state _directly_ against the local `csrfToken` (raw hex) — so after the cloud function is updated, those users would break.

`AuthManager.authWithWeb` now handles both formats:

```typescript
// Try base64 JSON first (v0.0.9+ cloud function)
const decoded = JSON.parse(Buffer.from(returnedState, 'base64').toString('utf8'));
if (typeof decoded?.csrf === 'string') {
  csrfFromState = decoded.csrf;
}
// Fallback: treat as raw hex token (≤v0.0.7 cloud function)
if (!csrfFromState) csrfFromState = returnedState;

if (!csrfFromState || csrfFromState !== csrfToken) {
  res.end('State mismatch. Possible CSRF attack.');
}
```

This means **both old and new cloud function versions work without a coordinated rollout.**

### 3. Tests

- `cloud_function/index.test.js` — 6 tests for state passthrough logic (verifies fix is correct, verifies old behaviour is gone)
- `workspace-server/src/__tests__/auth/AuthManager.test.ts` — 6 tests for CSRF extraction covering both format variants, null/empty state, and edge cases

## How to reproduce

1. Install the Google Workspace extension v0.0.9 in Claude Desktop / Cowork
2. Trigger any Google Workspace tool — the auth flow opens a browser
3. Authorize in Google
4. Observe: `State mismatch. Possible CSRF attack.` in the browser

## Workaround (until this is deployed)

```bash
node "/path/to/Claude Extensions/ant.dir.ant.0x55979e1.google-workspace/server/index.js" login
```

### DIFF
--- a/cloud_function/index.js
+++ b/cloud_function/index.js
@@ -111,9 +111,25 @@ async function handleCallback(req, res) {
           finalUrl.searchParams.append('token_type', token_type);
           finalUrl.searchParams.append('expiry_date', expiry_date.toString());
 
-          // SECURITY: Pass the CSRF token back to the client for validation.
-          if (payload.csrf) {
-            finalUrl.searchParams.append('state', payload.csrf);
+          // SECURITY: Pass the original base64-encoded state back to the client
+          // for validation. The client decodes it and extracts the `csrf` field itself.
+          //
+          // Previously, this code extracted only `payload.csrf` (a raw hex string) and
+          // returned it as `state`. That worked for workspace-server ≤ v0.0.7, which
+          // compared the returned value directly against the local csrf token.
+          //
+          // Starting with v0.0.9, the local server expects to receive the full base64
+          // JSON state back, then decodes it to extract the `csrf` field:
+          //
+          //   const decoded = JSON.parse(Buffer.from(returnedState, 'base64').toString('utf8'));
+          //   if (decoded.csrf !== localCsrfToken) → "State mismatch. Possible CSRF attack."
+          //
+          // Returning the raw hex here causes `JSON.parse` to fail, setting csrf to null
+          // and triggering a state mismatch on every auth attempt in v0.0.9+.
+          //
+          // Fix: return the original state parameter unchanged so the client can decode it.
+          if (state) {
+            finalUrl.searchParams.append('state', state);
           }
 
           return res.redirect(302, finalUrl.toString());
--- a/cloud_function/index.test.js
+++ b/cloud_function/index.test.js
@@ -1,0 +1,94 @@
+/**
+ * @license
+ * Copyright 2025 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * Tests for the OAuth cloud function — specifically the `state` passthrough
+ * fix introduced to support workspace-server v0.0.9+.
+ *
+ * The critical invariant: the cloud function must pass the original base64
+ * state back to the client unchanged, so the client can decode it and extract
+ * the csrf field for CSRF validation.
+ */
+
+// ---------------------------------------------------------------------------
+// Unit tests for the state passthrough logic (no HTTP server needed)
+// ---------------------------------------------------------------------------
+
+describe('state parameter passthrough in handleCallback', () => {
+  /**
+   * Mirrors the state-handling block in handleCallback to test it in isolation.
+   * Returns the value that would be appended as the `state` query param on the
+   * redirect URL, or null if no state would be appended.
+   */
+  function buildRedirectState(stateParam) {
+    if (!stateParam) return null;
+    if (stateParam.length > 4096) throw new Error('State parameter exceeds size limit of 4KB.');
+
+    const payload = JSON.parse(Buffer.from(stateParam, 'base64').toString('utf8'));
+
+    if (payload && payload.manual === false && payload.uri) {
+      // The fix: return `stateParam` unchanged, NOT `payload.csrf`
+      return stateParam;
+    }
+    return null;
+  }
+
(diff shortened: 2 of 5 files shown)


allenhutchison:
Hi @bs-nubank — thank you for the very thorough write-up; the repro steps and the protocol drift analysis are exceptionally well done. Before we move forward I want to flag something I want your help understanding, because what you're describing doesn't seem to match the source in this repo.

**The repo's source on `main` (and on `v0.0.7`) doesn't decode base64 JSON on the receive side.** `AuthManager.authWithWeb` reads `state` directly off the query string and compares it as a raw string against `csrfToken`:

```ts
// workspace-server/src/auth/AuthManager.ts
const returnedState = qs.get('state');
if (returnedState !== csrfToken) {
  res.end('State mismatch. Possible CSRF attack.');
}
```

That has been the receive-side logic since the initial commit — `git log -S "Buffer.from"` on `AuthManager.ts` shows the only base64 work happens on the *send* side (encoding `{uri, manual, csrf}` to send *to* Google), and `git log -S "JSON.parse"` on the same file turns up nothing on the receive side prior to your PR.

So today's flow is:
1. Extension sends `state = base64(JSON({uri, manual, csrf}))`
2. Cloud function decodes it, redirects to `payload.uri` with `state=payload.csrf` (raw hex)
3. Extension does direct equality of returned hex against `csrfToken`

That's internally consistent and works, which is why I think `v0.0.7` and earlier users aren't hitting this.

**Where I'm stuck:** the snippet you quoted as "v0.0.9 — extension/server/index.js" (the minified `JSON.parse(Buffer.from(i, 'base64')...).csrf` block) doesn't correspond to anything I can find in this repo's history. There's no `v0.0.9` tag here either — the highest release tag is `v0.0.7`, and `package.json` is `0.0.7`. Your repro mentions Claude Desktop / Cowork, which is a different distribution channel than what we publish (this is the Gemini CLI extension).

**Could you point me at the exact source for that v0.0.9 file?** I want to make sure we're fixing the right thing. My current concern is: if "v0.0.9" is a downstream redistribution (e.g. a Claude Desktop bundle that someone repackaged), the fix likely belongs there, because changing this repo's cloud function to return full base64 JSON would break every currently-deployed `v0.0.7` user — their AuthManager does direct hex equality and would fail on every auth from the moment the cloud function rolls out, and the backward-compat fallback you added only helps the *next* workspace-server release, not the installs already in the field.

chandanchowdhury:
Interesting timing! 

I was trying to get the extension working **without** using the Cloud Function and stumbled on the same issue.

**Background**: Our GCP `OAuth consent screen` -> `Audience` -> `User Type` is set to `Internal` so we are okay to share the OAuth Client ID and Secret to our users as the client will not be usable outside our org. This approach removes any possibility of us as the Cloud Function operator from ever being able to see a user's OAuth token.

**Summary**: In absence of the _proxy_ (cloud function) the response goes directly from Google to the `/oauth2callback` and the response has the original [`state` containing `csrf`](https://github.com/gemini-cli-extensions/workspace/blob/main/workspace-server/src/auth/AuthManager.ts#L319). 
Later the code at [line-356](https://github.com/gemini-cli-extensions/workspace/blob/main/workspace-server/src/auth/AuthManager.ts#L356) is checking if `state` is the `csrf` we sent and is failing as Google returns the original `state` which was base64 encoded JSON.

It would be more consistent and secure for the code in `AuthManager.ts` if the cloud function returns the `state` exactly as it was sent (similar to what Google auth server does).

Sum1cares (approved):

Sum1cares (approved):

Sum1cares (approved):

allenhutchison:
Closing due to lack of response to my comments from April. Feel free to reopen if you want to continue to pursue this.

PR #348, opened 2026-04-25
feat(slides,drive): add createFromJson, drive primitives, and theme system

n0012:
## What this adds

Five new tools for building Google Slides presentations programmatically, plus three new Drive primitives for safe-by-default image staging.

---

### `slides.createFromJson` — blueprint-to-slides in one call

Callers describe a deck as a JSON blueprint; the server translates it into a Slides API `batchUpdate`. No knowledge of raw API shape required.

**Color aliases** — named colors, never RGB:
| Alias | Value | Use |
|---|---|---|
| `text` | #202124 near-black | body text |
| `primary` | #101828 dark | dark backgrounds |
| `primary_text` | #FFFFFF white | text on dark |
| `blue` | #1A73E8 Google Blue | accents, labels |
| `red/yellow/green` | Google brand colors | brand bar |
| `surface` | #F1F3F4 light gray | card backgrounds |
| `text_muted` | #757575 gray | secondary text |

**Theme system** — 12 named themes (`google`, `exec`, `pitch`, `technical`, `workshop`, `dark`, `demo`, `hcls`, `customer`, `simple`, `google-dark`, `google-minimal`) drive font family, accent color, and layout guidance in the tool description.

**Speaker notes** — include `"speaker_notes"` in each slide object and they're written automatically. Tool warns when notes are missing and requests a second pass.

**Layer ordering** — elements render shapes → images → text, then by `layer` value. Background shapes reliably appear behind text without manual sequencing.

**Blueprint format:**
```json
{ "slides": [
    { "speaker_notes": "...", "elements": [...] },
    { "speaker_notes": "...", "elements": [...] }
  ]
}
```

**Element schema:** `type` (`text` | `shape` | `image`), `position` (`{x,y,w,h}` in points on 720×405 canvas), `layer` (z-index), `content`, `url`, and a `style` object with: `size`, `bold`, `color`, `bg_color`, `no_border`, `align`, `vertical_align`.

---

### Drive primitives — split for safe-by-default uploads

The Slides API's `createImage` endpoint requires a publicly accessible URL (per [Google's docs](https://developers.google.com/workspace/slides/api/guides/add-image)) — OAuth tokens in URLs are not honored. To support image-heavy workflows without making the upload tool itself dangerous, the share lifecycle is split across three tools:

- **`drive.uploadFile`** — uploads a local file to Drive. **File is PRIVATE by default** (no share granted). Returns `id`, `name`, `webViewLink`.
- **`drive.addPublicAccess`** — explicit opt-in: grants `anyone:reader` on an existing file, returns the public `imageUrl` and the permission ID. Surfaces Workspace `publishOutNotPermitted` clearly so callers can fall back to another hosting path (GCS signed URLs, etc.).
- **`drive.removePublicAccess`** — revokes every `anyone:*` permission on a file. Idempotent. File stays in Drive — only the public link is closed.

Typical use for embedding a local image in a slide:

```
1. id       = drive.uploadFile(localPath)            # private
2. imageUrl = drive.addPublicAccess(id)              # explicit opt-in
3. slides.createFromJson(..., url: imageUrl, ...)
4. drive.removePublicAccess(id)                       # close the window
```

---

### Supporting tools

- **`slides.create`** — create a blank presentation, returns `{presentationId, url}`
- **`slides.batchUpdate`** — raw Slides API request array passthrough
- **`slides.getText` / `getMetadata` / `getImages` / `getSlideThumbnail`** — read tools
- **`slides.getSpeakerNotes` / `updateSpeakerNotes`** — read and write speaker notes per slide

---

## Design notes

- All new tools registered via `server.registerTool` and gated through `feature-config.ts`. `slides.write` and `drive.write` groups carry the new tools with correct scope requirements.
- Color alias system replaces raw RGB so output stays consistent with the active theme.
- Speaker notes written inline from blueprint (no second pass required unless omitted).
- Drive upload split into three primitives so the safe default is private; sharing is always explicit and reversible.

## Validation

- TypeScript build passes (`npm run build`)
- Full round-trip exercised end-to-end on both personal and corporate Google accounts: presentation creation, multi-slide blueprints with shapes/text/images, color aliases, speaker notes, image upload + share + revoke.
- On corporate Workspace domains where `publishOutNotPermitted` blocks `addPublicAccess`, the error is clearly surfaced so callers can route images through an alternative public host.


n0012:
test with `gemini extensions install https://github.com/n0012/workspace`

new tool that can batch create slides is slides.createFromJson

n0012:
## Bug fix bundled in this PR: `docs.getText` field mask error

While testing this PR I hit a systematic breakage in `docs_getText` (and `writeText`/`replaceText`) affecting all Google Docs — including single-tab docs with no suggestions. Root cause and fix:

**Root cause:** `docs.documents.get` with `includeTabsContent: true` and a wildcard field mask (`tabs`, `tabs.documentTab.body`, etc.) causes the API to validate the mask as including suggestion/comment sub-fields (`suggestedInsertionIds`, `suggestedParagraphStyleChanges`, etc.). The API then rejects with:
```
Field mask cannot retrieve comment-specific fields when include_comments is false.
```
A previous attempt added `suggestionsViewMode: 'PREVIEW_WITHOUT_SUGGESTIONS'` to suppress suggestion data, but this doesn't help — the field mask is validated **before** the view-mode filter is applied.

**Fix (commit 02b3502):** Replaced all three `documents.get` calls in `DocsService` with explicit field masks (`DOCS_READ_FIELDS`, `DOCS_END_INDEX_FIELDS`) that enumerate only the fields `_readStructuralElement` actually reads — no suggestion or comment sub-fields anywhere in the tree. Handles up to 3 levels of tab nesting (Google Docs maximum) and one level of table nesting.

n0012:
@allenhutchison — would appreciate a review when you get a chance! This adds `slides.createFromJson`, `slides.insertImageSlide`, and `drive.uploadFile` — tools we've been using in production with Claude Code + Gemini CLI for building AI-generated slide decks. CLA is green. Happy to address any feedback.

Sum1cares (approved):

Sum1cares (approved):

Sum1cares (approved):

Sum1cares (approved):

Sum1cares (approved):

workspace-server/src/index.ts
```diff
@@ -476,6 +476,197 @@ async function main() {
     slidesService.getSlideThumbnail,
   );
 
+  // Speaker notes tools — approach adapted from PR #235
+  // https://github.com/gemini-cli-extensions/workspace/pull/235 by @stefanoamorelli
+  server.registerTool(
```
allenhutchison: **[blocking]** These five new tools (this one plus `slides.updateSpeakerNotes` at 495, `slides.create` at 515, `slides.batchUpdate` at 527, and `slides.createFromJson` at 652) call `server.registerTool(...)` directly, but every other tool in this file goes through the wrapped `registerTool(...)` helper defined at lines 169–182. The wrapper is what honors `enabledTools` from feature-config — without it, `slides.write` will register regardless of `defaultEnabled: false`. Should be a `s/server.registerTool/registerTool/` on all five.

workspace-server/src/services/SlidesService.ts
```diff
@@ -280,6 +356,699 @@ export class SlidesService {
...
+            requestBody: { requests: noteRequests },
+          });
+          logToFile(
+            `[SlidesService] Wrote speaker notes for ${notesSlides.length} slides`,
+          );
+        }
+      }
+
+      // Delete the default blank slide ("p") that Google creates with new presentations
+      try {
```
allenhutchison: **[blocking]** Two concerns on this delete-`'p'` block: (1) `createFromJson` takes an arbitrary `presentationId`, so if a caller appends to an existing deck and `'p'` happens to be a real slide, we silently delete it; (2) the `catch {}` at 869 swallows everything — auth revocation, quota, 5xx — with no log.

Could we gate the delete on an explicit `isNewPresentation` flag (or fetch the slide list first and only delete if `'p'` exists), and at minimum log unexpected catches? Silent catches are something we try to avoid repo-wide.

workspace-server/src/services/DriveService.ts
```diff
@@ -639,4 +639,137 @@ export class DriveService {
...
+            text: JSON.stringify({
+              id,
+              permissionId: perm.data.id,
+              imageUrl,
+            }),
+          },
+        ],
+      };
+    } catch (error) {
+      return this.handleError('drive.addPublicAccess', error);
```
allenhutchison, line 775: **[blocking]** The PR description calls out `publishOutNotPermitted` as a feature surface, but here we fall through to the generic `handleError` that just stringifies `error.message` — the caller gets an opaque "insufficient permissions" string with no signal that it's org policy. Could we detect `error.errors?.[0]?.reason === 'publishOutNotPermitted'` (and `cannotShareOutsideDomain`) before the generic path and return a structured response with an actionable hint?

workspace-server/src/services/SlidesService.ts
```diff
@@ -280,6 +356,699 @@ export class SlidesService {
...
+                contentAlignment:
+                  style.vertical_align as slides_v1.Schema$ShapeProperties['contentAlignment'],
+              },
+              fields: 'contentAlignment',
+            },
+          });
+        }
+
+        // Bold phrases
+        if (style.bold_phrases) {
```
allenhutchison, line 1673: **[blocking]** `content.indexOf("", n)` returns `n` for every position, so an empty `phrase` runs `content.length + 1` times and emits invalid `updateTextStyle` requests with `startIndex === endIndex` — Slides API rejects the whole batchUpdate. Same shape applies to the `links` loop further down. Either `if (!phrase) continue;` at the top of each loop, or add `.min(1)` to the schema.

workspace-server/src/services/SlidesService.ts
```diff
@@ -13,6 +13,82 @@ import { logToFile } from '../utils/logger';
...
+  background: RGB;    // Slide background
+  fontFamily: string; // Default font family
+  // Extended palette — optional, used when alias is referenced
+  accent1?: RGB;
+  accent2?: RGB;
+  accent3?: RGB;
+  accent4?: RGB;
+}
+
+const THEMES: Record<string, Theme> = {
```
allenhutchison: **[blocking]** The PR description advertises 12 themes (`exec`, `pitch`, `technical`, `workshop`, `dark`, `demo`, `hcls`, `customer`, `simple`, `google-dark`, `google-minimal`) but I only see `google` defined here. `THEMES['exec']` returns `undefined` and `createFromJson` hardcodes `THEMES['google']` so the others are unreachable anyway. Either ship the missing themes or pare the description and tool docs back to what actually exists. Also are we sure we want to ship a "Google" theme in a public extension?

workspace-server/src/services/SlidesService.ts
```diff
@@ -13,6 +13,82 @@ import { logToFile } from '../utils/logger';
...
+ * Returns undefined if alias unknown or no theme active.
+ */
+function resolveColor(color: ColorValue | undefined, theme: Theme): RGB | undefined {
+  if (!color) return undefined;
+  if (typeof color !== 'string') return color as RGB;
+  const key = COLOR_ALIASES[color.toLowerCase()];
+  return key ? (theme[key] as RGB) : undefined;
+}
+
 export class SlidesService {
```
allenhutchison, line 128: **[blocking]** No tests in this PR — flagging because our convention is that every service has unit tests (`SlidesService.test.ts` is ~400 lines, `DriveService.test.ts` is ~1,400). The googleapis/AuthManager mocking harness is already wired up so the marginal cost is low.

Priority targets:
- `resolveColor` — pure function, several alias paths
- `removePublicAccess` — filters on `type === 'anyone'`; a regression here is a security bug
- `uploadFile` missing-file path
- One blueprint-translation snapshot for `createFromJson`

Let me know if you'd like a starter test to anchor the pattern.

workspace-server/src/services/DriveService.ts
```diff
@@ -639,4 +639,137 @@ export class DriveService {
...
+    try {
+      const drive = await this.getDriveClient();
+      const id = extractDocumentId(fileId);
+      const perm = await drive.permissions.create({
+        fileId: id,
+        supportsAllDrives: true,
+        requestBody: { role: 'reader', type: 'anyone' },
+        fields: 'id',
+      });
+      const imageUrl = `https://drive.google.com/uc?export=download&id=${id}`;
```
allenhutchison: **[important]** Did you test this returned `imageUrl` end-to-end with `slides.createImage`? The `drive.google.com/uc?export=download` form is the legacy endpoint and returns an HTML interstitial for files >~25MB, which the Slides API server-side fetch won't follow. If you've confirmed it works in your manual testing, ignore me — otherwise `https://lh3.googleusercontent.com/d/<id>` is more reliable.

workspace-server/src/index.ts
```diff
@@ -476,6 +476,197 @@ async function main() {
...
+          .optional()
+          .describe('Border weight in points.'),
+        no_border: z
+          .boolean()
+          .optional()
+          .describe('Remove border from shape.'),
+        font_family: z
+          .string()
+          .optional()
+          .describe('Font family name (e.g. "Arial", "Roboto"). Defaults to "Arial".'),
```
allenhutchison: **[important]** Description says `font_family` defaults to `"Arial"`, but the implementation at `SlidesService.ts:648` falls back to `theme.fontFamily` (`"Google Sans"`). Could we update this to `Defaults to the active theme font ("Google Sans").`?

workspace-server/src/index.ts
```diff
@@ -476,6 +476,197 @@ async function main() {
...
...des should be mostly white/background with dark text. Use colored elements sparingly — a thin accent line, a highlighted key metric, a section label. Not every slide needs a colored header bar. Whitespace IS the design.\n\nTECHNICAL NOTES:\n- Layers: lower values render first (backgrounds=0, boxes=1, text=2+). Missing layers cause text to be hidden behind shapes.\n- Font sizes: titles ~20-24pt bold, subheadings ~12-14pt, body ~10-12pt.\n- Text boxes clip silently — size h generously.\n\nSTYLE PROPERTIES: size, bold, italic, underline, strikethrough, align (START|CENTER|END), vertical_align (TOP|MIDDLE|BOTTOM), indent, color, bg_color, border_color, border_weight, no_border, font_family ("theme" to inherit theme font), bold_phrases, bold_until, links ([{text,url}]).\n\nCOLOR ALIASES: "primary" (#202124 near-black), "primary_text" (white), "secondary" (#1A73E8 Blue 600), "text" (#1F1F1F), "text_muted" (#444746), "surface" (Blue 50), "surface_alt" (Green 50), "background" (white). Brand colors: "blue" (#4285F4), "red" (#EA4335), "yellow" (#FBBC05), "green" (#34A853). OR use RGB 0-1 objects for one-off colors. Image URLs with unresolved placeholders are replaced with a fallback icon.',
```
allenhutchison: **[important]** The schema/description uses `MUST` and `REQUIRED` for `speaker_notes`, but the code at `SlidesService.ts` treats it as optional and returns an `action_required` hint when missing. Mismatch is misleading to MCP clients that read the schema as authoritative — could we either tighten the schema (reject without notes) or soften the wording to "strongly recommended"?

workspace-server/src/index.ts
```diff
@@ -476,6 +476,197 @@ async function main() {
...
+          .optional()
+          .describe('Vertical content alignment.'),
+        color: z
+          .object({
+            red: z.number(),
+            green: z.number(),
+            blue: z.number(),
+          })
+          .optional()
+          .describe('Text color (RGB 0-1).'),
```
allenhutchison: **[important]** This color schema only accepts `{red, green, blue}` objects, but the alias system in `SlidesService.ts:60` expects string keys (`"primary"`, `"blue"`, etc.) — meaning the whole color-alias feature is unreachable from the public tool boundary even though the tool description heavily promotes it. Could the schema accept `RGB | string` (ideally a literal union derived from the alias keys so callers get autocomplete)? Same shape applies to `bg_color` and `border_color` below.

workspace-server/src/services/DriveService.ts
```diff
@@ -639,4 +639,137 @@ export class DriveService {
...
+      const anyonePerms = (perms.data.permissions ?? []).filter(
+        (p) => p.type === 'anyone',
+      );
+      if (anyonePerms.length === 0) {
+        return {
+          content: [{ type: 'text' as const, text: JSON.stringify({ id, removed: [] }) }],
+        };
+      }
+      const removed: string[] = [];
+      for (const p of anyonePerms) {
```
allenhutchison, line 799: **[important]** If permission #3 of 5 fails mid-loop (race, revoked auth), we throw and `removed` never gets returned — caller can't tell which perms were actually deleted, which breaks the "idempotent" contract on retry. Could each `delete` be in its own try/catch, collecting `{removed, failed}` in the response?

workspace-server/src/services/SlidesService.ts
```diff
@@ -280,6 +356,699 @@ export class SlidesService {
...
+              fields: fields.join(','),
+            },
+          });
+        }
+      } else if (el.type === 'image') {
+        const objId = getId('img');
+        // Sanitize URLs that contain unresolved template placeholders (e.g. from LLM output)
+        let imageUrl = el.url ?? '';
+        if (imageUrl.includes('{') || imageUrl.includes('%7B')) {
+          imageUrl = 'https://img.icons8.com/m_rounded/512/4285F4/info.png';
```
allenhutchison, line 1553: **[important]** Substituting an icons8 icon when the URL contains `{` is a reasonable fallback, but please add a `warnings: [{slideIndex, elementIndex, issue: "unresolved url placeholder, substituted fallback"}]` entry on the response so the caller knows it happened. Right now a malformed blueprint silently fills the deck with info icons.

workspace-server/src/services/SlidesService.ts
```diff
@@ -280,6 +356,699 @@ export class SlidesService {
...
+      const errorMessage =
+        error instanceof Error ? error.message : String(error);
+      logToFile(
+        `[SlidesService] Error during slides.create: ${errorMessage}`,
+      );
+      return {
+        content: [
+          {
+            type: 'text' as const,
+            text: JSON.stringify({ error: errorMessage }),
```
allenhutchison, line 1356: **[important]** Error returns from the new SlidesService methods (this `create` plus `batchUpdate` and `createFromJson`) don't set `isError: true` on the MCP response — `DriveService.handleError` does this correctly at line 47. MCP clients use that flag to distinguish failure from success, so without it these look like successful responses that happen to contain an `error` string.

workspace-server/src/services/SlidesService.ts
```diff
@@ -280,6 +356,699 @@ export class SlidesService {
...
+    theme: Theme,
+  ): slides_v1.Schema$Request[] {
+    const requests: slides_v1.Schema$Request[] = [];
+
+    const getId = (prefix: string) => {
+      objCounter.value += 1;
+      return `${prefix}_${Date.now()}_${objCounter.value}`;
+    };
+
+    // Sort: shapes first, then images, then text; within each group by layer
```
allenhutchison: **[nit]** Comment says "shapes first, then images, then text", but the sort key `layerVal * 10 + typeVal` puts layer first with type as the tiebreaker — the opposite. The behavior matches the tool description's "lower layers render first" promise, so the code is right; just the inline comment is misleading.

workspace-server/src/services/SlidesService.ts
```diff
@@ -280,6 +356,699 @@ export class SlidesService {
...
+                  },
+                  fields: 'bold',
+                },
+              });
+              searchFrom = idx + 1;
+            }
+          }
+        }
+
+        // Bold until (legacy)
```
allenhutchison: **[nit]** `// Bold until (legacy)` — calling it "legacy" is confusing since it's brand new in this PR. Drop the tag, or explain why it's deprecated on arrival?

allenhutchison (changes requested):

allenhutchison:
Hi @n0012 👋 — just checking in on this one. This is a genuinely impressive contribution: `createFromJson`, the Drive staging primitives, and the 12-theme system are a lot of thoughtful work, and the blueprint-to-slides approach is really nice.

I left a round of review on May 26 — 15 inline comments, mostly focused things rather than anything fundamental. No rush at all, but I wanted to make sure it didn't slip by unnoticed since it landed just after a batch of approvals.

Are you planning to keep going with it? If so, I'm happy to help work through any of the comments or clarify anything that's unclear — just let me know. And if you'd like a second pass on anything once you've pushed updates, tag me and I'll re-review promptly. Thanks again for this! 🙌

bdoyle0182:
@n0012 if planning to continue with this one, please refactor around the baseline slides tools that just got merged into the main branch!

n0012:
Thanks, had some time away but will review again and apply feedback per these comments.  Please keep an eye out @allenhutchison and @bdoyle0182

n0012:
Thanks @allenhutchison and @bdoyle0182 for the thorough review, and @gemini-code-assist for the automated pass. I've had some time away but have now worked through **every** comment. Summary below, grouped by reviewer. The branch is rebased on the current `main` and sits at `5cb49dc`.

## @allenhutchison — blocking

- **`registerTool` wrapper (the 5 tools at index.ts:481/495/515/527/652)** — ✅ All slides write tools now register through the wrapped `registerTool(...)` helper; there are **no** remaining `server.registerTool(...)` calls. `slides.write` now correctly honors `defaultEnabled:false` and `WORKSPACE_FEATURE_OVERRIDES`.
- **Silent delete of slide `'p'`** — ✅ The delete is gated behind an explicit `isNewPresentation` flag (default `false`), and even then only fires after fetching the slide list and confirming `'p'` actually exists. The previously-silent `catch` now logs. Appending to an existing deck can never delete a real slide.
- **`publishOutNotPermitted` opaque error** — ✅ `DriveService.addPublicAccess` now detects `reason === 'publishOutNotPermitted' || 'cannotShareOutsideDomain'` and returns a structured, actionable hint before falling through to the generic handler.
- **Empty-phrase `indexOf` bug** — ✅ `if (!phrase) continue;` / `if (!linkDef.text) continue;` guard both the `bold_phrases` and `links` loops, in **both** `createFromJson` and `setText`.
- **12 themes vs 1 / "ship a Google theme in a public extension?"** — ✅ Resolved by removing branding from the server entirely. It now ships a **single neutral palette** (`Arial`), and the `theme` selector param is gone. Color aliases resolve against that neutral palette. Google branding (Google Sans + brand bar) lives in the consuming skill, not the extension — which I think directly answers the public-extension concern.
- **No tests** — ✅ Added `SlidesService.test.ts` (+438) and `DriveService.test.ts` (+135) covering the targets you named: `resolveColor` alias paths, `removePublicAccess` (incl. the `type === 'anyone'` filter), `uploadFile` missing-file, and a `createFromJson` blueprint-translation case.

## @allenhutchison — important

- **lh3 image URL** — ✅ `addPublicAccess` now returns `https://lh3.googleusercontent.com/d/<id>` instead of the legacy `uc?export=download` form.
- **`font_family` default doc** — ✅ Reworded to "defaults to the active theme font."
- **`speaker_notes` MUST/REQUIRED mismatch** — ✅ Softened to "strongly recommended"; the code returns the `action_required` hint when notes are omitted, so schema and behavior now agree.
- **Color schema only accepted RGB** — ✅ `slidesColorSchema = z.union([z.string(), {red,green,blue}])`, applied to `color`/`bg_color`/`border_color`, so the alias system is reachable from the tool boundary.
- **`removePublicAccess` partial-failure** — ✅ Each delete is isolated in its own try/catch; the response returns `{ removed, failed }` so retries stay idempotent.
- **Silent icon fallback on placeholder URLs** — ✅ Now emits a `warnings` entry on the response (and documented in the tool description).
- **Missing `isError: true`** — ✅ The new SlidesService error returns now set `isError: true`.

## @allenhutchison — nits
- ✅ Sort comment reworded (layer is primary; type is the tiebreaker within a layer).
- ✅ Dropped the misleading "legacy" tag on `bold_until`.

## @gemini-code-assist
- ✅ `registerTool` wrapper on all flagged tools (covered above).
- ✅ `batchUpdate.requests` accepts `string | array`.
- ✅ `slideJson` accepts the structured `slideElementSchema` union (or a JSON string).
- ✅ `createFromJson` defaults a missing `elements` to `[]` — no crash on `{ "slides": [{}] }`.
- ✅ `createFromJson` omits `insertionIndex` so slides append to the end.

## @bdoyle0182
- ✅ Rebased onto current `main`. The PR builds on the merged baseline slides tools (`addSlide`/`addShape`/`addImage`/`addTable`) rather than redefining them — the net-new surface here is `createFromJson`, `setText`, `batchUpdate`, and the `drive.*` staging primitives.

Would really appreciate another pass when you have a chance, @allenhutchison @bdoyle0182 — happy to keep iterating on anything that's still off.

n0012:
Hi team — gentle nudge on this one: have you had a chance to review? Happy to hop on a call to walk through it, or to split it into smaller PRs if that would make review easier. Thanks!

workspace-server/src/index.ts
```diff
@@ -893,6 +952,210 @@ async function main() {
...
+
+  const slideObjectSchema = z.object({
+    elements: z.array(slideElementSchema),
+    speaker_notes: z.string().optional(),
+  });
+
+  registerTool(
+    'slides.createFromJson',
+    {
+      description:
```
bdoyle0182, line 1056: This is a really long description for a tool.

A lot of this imo is not related to what would go in tool description, but how to use the tool in a specific way through progressive disclosure in a skill

workspace-server/src/index.ts
```diff
@@ -379,6 +419,25 @@ async function main() {
     docsService.getText,
   );
 
+  registerTool(
+    'docs.getInlineImages',
+    {
+      description:
+        "Returns every inline image in a Google Doc, in document order, as an array of { objectId, contentUri, altTitle, altDescription }. The contentUri is a temporary googleusercontent.com URL fetchable by other Google services (e.g. the Slides API image fetcher) WITHOUT public link sharing on the source Doc. contentUri is short-lived — consume it promptly.",
```
bdoyle0182, line 426: description can be shortened significantly. model should already know it's working with google docs by the tool name

"in document order" i would expect the model to be able to infer this

"(e.g. the Slides API image fetcher)" not relevant to tool description.

"contentUri is short-lived — consume it promptly." i have concerns with this tool's inclusion if it's functionality is time based.

workspace-server/src/index.ts
```diff
@@ -330,6 +330,46 @@ async function main() {
...
+      },
+    },
+    docsService.createFromMarkdown,
+  );
+
+  registerTool(
+    'docs.updateFromMarkdown',
+    {
+      description:
+        "Replaces an existing Google Doc's entire content from a Markdown string (same clean Markdown→Doc conversion), keeping the same document ID and link. Use to refresh a doc previously created from Markdown.",
```
bdoyle0182, line 358: "same clean Markdown->Doc conversion" - same to what? the agent may not get the create tool in the client's semantic tool search.

workspace-server/src/index.ts
```diff
@@ -330,6 +330,46 @@ async function main() {
     docsService.writeText,
   );
 
+  registerTool(
```
bdoyle0182, line 333: create and update should probably be one tool with an update flag as a parameter

bdoyle0182:
I think the area of concern is too broad on this. Each thing being added should be its own PR i.e. docs markdown tools should be one PR

n0012:
Thanks again @allenhutchison and @bdoyle0182 for the thorough review — really appreciate the time you put in.

I've now worked through all of the feedback in my fork:

- **Blocking/important items** — `registerTool` wrapper on all slides write tools, `isNewPresentation` gating on the blank-slide delete, empty-phrase guards, `isError: true` on error returns, `publishOutNotPermitted`/`cannotShareOutsideDomain` hints, per-permission partial-failure handling in `removePublicAccess`, `string | RGB` color-alias schema, `lh3.googleusercontent.com` image URLs, placeholder-URL `warnings`, single neutral palette (no bundled branding), and service tests.
- **Design notes** — merged `docs.createFromMarkdown` + `docs.updateFromMarkdown` into a single `docs.fromMarkdown` (create vs. overwrite driven by `documentId`), and trimmed the `createFromJson` and `getInlineImages` descriptions down to the tool contract.

Rather than reopen this large PR, I'll bring the changes in as smaller, focused PRs so they're easier to review. Thanks again for all the input!

PR #349, opened 2026-04-29
fix(docs): resolve docs.getText API errors

allenhutchison:
This PR fixes a "comment-specific fields" error in `docs.getText` and tightens the `documents.get` calls in DocsService.

## Changes

1. **Refine the tabs field mask** from the broad `'tabs'` (or `'title,tabs'`) to a specific selection — `tabs(tabProperties,documentTab(body,headers,footers,footnotes))` — so it doesn't trigger internal API conflicts when combined with `includeTabsContent: true`. Applied at all three `documents.get` call sites in `DocsService.ts`, extracted to a single exported `TABS_FIELD_MASK` constant.
2. **Add `suggestionsViewMode: 'PREVIEW_WITHOUT_SUGGESTIONS'`** to the `getText` call so unaccepted suggestion markup doesn't leak into the extracted plain text.
3. **Update SKILL.md** to point users at `drive.getComments` instead of the (non-existent) `docs.getComments`.

## Tests

- Existing `replaceText` tests updated to reference the new `TABS_FIELD_MASK` constant.
- New assertion in the first `getText` test pinning the refined `documents.get` args (field mask + `suggestionsViewMode`) so future regressions are caught at PR time rather than only via API errors against real Docs.

## Drive-by: unbreak `print-scopes.ts` under TS 6

Rolled in here because it's blocking CI on `main` itself after #344 (TS 6 bump) and #347 (drift-guard test) landed in the same window:

- TS 6 became stricter about implicit `rootDir`. ts-node compiling `scripts/print-scopes.ts` against the root tsconfig (which only includes `workspace-server/src/`) failed with TS5011 — the file is outside the included tree, so TS couldn't infer rootDir.
- Added a small `scripts/tsconfig.json` with explicit `rootDir: ".."`, and pass `--project scripts/tsconfig.json` from both call sites (the drift-guard test and `setup-gcp.sh`).

This unblocks `main`'s CI as a side effect of merging this PR.

### DIFF
--- a/scripts/setup-gcp.sh
+++ b/scripts/setup-gcp.sh
@@ -95,7 +95,7 @@ echo ""
 # workspace-server/src/features/feature-config.ts. See issue #323.
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
-SCOPES_OUTPUT=$(cd "$REPO_ROOT" && npx --no-install ts-node --transpile-only scripts/print-scopes.ts 2>&1)
+SCOPES_OUTPUT=$(cd "$REPO_ROOT" && npx --no-install ts-node --transpile-only --project scripts/tsconfig.json scripts/print-scopes.ts 2>&1)
 if [ $? -ne 0 ]; then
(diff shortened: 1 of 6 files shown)


abhipatel12 (approved):
LGTM!

PR #382, opened 2026-06-01
fix(docs): drop tabs field mask to support smart-chip docs in get/write/replace

elisabuonanni-nu:
## Summary

`docs_getText`, `docs_writeText`, and `docs_replaceText` all fail against any Google Doc containing smart chips (Person, Date, RichLink, Dropdown, etc.) with:

> Field mask cannot retrieve comment-specific fields when include_comments is false.

The Docs API treats some content reachable via the `fields: 'tabs'` mask as comment-anchored once smart chips are present in the doc. The REST API does **not** expose any `include_comments` parameter that callers can set — an attempt to pass `includeComments: true` is rejected by the SDK as `Unknown name 'include_comments'`. The fix is to drop the explicit field mask while keeping `includeTabsContent: true`, which still returns the tab content but no longer trips the trigger.

## What changed

`workspace-server/src/services/DocsService.ts`, three call sites:

- `writeText` — end-of-doc preflight (resolving `position: 'end'`)
- `getText`
- `replaceText` — preflight that locates ranges before issuing `replaceAllText` requests

Each `docs.documents.get(...)` call now passes only `documentId` and `includeTabsContent: true`. The downstream code only consumes `tabs.tabProperties.tabId` and `tabs.documentTab.body.content` (via the existing `_flattenTabs` helper), all of which are returned in the API default field set.

## Repro

Any Google Doc with at least one smart chip and any of the three affected tools. Before this PR: field-mask error. After: full body returned, smart chips preserved as inline tokens.

## Test plan

- [x] `npm run build` clean
- [x] `docs_getText` against a doc with Person, Date, and RichLink chips → returns body
- [x] No regression on plain docs (still returns content via the default mask)
- [ ] Reviewer: confirm jest suite still green (I did not run the full suite locally)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/workspace-server/src/services/DocsService.ts
+++ b/workspace-server/src/services/DocsService.ts
@@ -300,10 +300,13 @@ export class DocsService {
         if (position === 'beginning') {
           index = 1;
         } else if (position === 'end') {
-          // Discover the end index by reading the document (required for tabs)
+          // Discover the end index by reading the document (required for tabs).
+          // No explicit fields mask: when the doc contains smart chips the
+          // server rejects a `tabs` mask as implicitly comment-anchored.
+          // Letting the API return its default field set works on both
+          // smart-chip and plain docs.
           const res = await docs.documents.get({
             documentId: id,
-            fields: 'tabs',
             includeTabsContent: true,
           });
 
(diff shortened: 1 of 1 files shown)


allenhutchison:
@elisabuonanni-nu you'll need to agree to the CLA before we can begin work on this PR. Thanks.

allenhutchison:
Closing this PR for now due to lack of activity.

PR #383, opened 2026-06-02
fix(calendar): include attachments in listEvents field mask

euclides-filho:
## What

`listEvents` requests an explicit `fields` mask, and that mask omitted `attachments`. Because `events.list` only returns fields named in the mask, native event attachments were silently dropped from every `listEvents` result.

This adds `attachments(fileId,fileUrl,title,mimeType,iconLink)` to the mask.

## Why this is a consistency fix, not a new feature

Every *other* calendar method in `CalendarService` already exposes/handles attachments — `listEvents` was the lone exception:

- **`getEvent`** calls `events.get` with **no field mask**, so it already returns attachments.
- **`createEvent` / `updateEvent`** fully support attachments via the `EventAttachment` interface and `applyMeetAndAttachments` (`supportsAttachments`).

So a caller can create an event with an attachment and read it back via `getEvent`, but the same attachment vanishes when the event is fetched through `listEvents`. This change aligns `listEvents` with the rest of the service.

Note: reading attachments needs **no** `supportsAttachments` parameter — that flag is write-only (insert/update/patch). Adding the field to the read mask is sufficient.

## Impact

- Purely additive to the `fields` mask; no change to any existing field or to request/response shapes beyond the new `attachments` array.
- Slightly larger `listEvents` payloads when events carry attachments.

## Testing

- New unit test: asserts the `events.list` field mask includes `attachments(` and that an attachment's `fileId` survives into the tool payload.
- Updated the two existing field-mask assertions to include the new field.
- Full suite green: **512 passed, 28 suites**.

### DIFF
--- a/workspace-server/src/__tests__/services/CalendarService.test.ts
+++ b/workspace-server/src/__tests__/services/CalendarService.test.ts
@@ -515,7 +515,9 @@ describe('CalendarService', () => {
       expect(callArgs.fields).toContain('attachments(');
 
       const returned = JSON.parse(result.content[0].text);
-      expect(returned[0].attachments[0].fileId).toBe('example-attachment-file-id');
+      expect(returned[0].attachments[0].fileId).toBe(
+        'example-attachment-file-id',
+      );
     });
 
     it('should list events with a default timeMax', async () => {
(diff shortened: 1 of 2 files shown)


allenhutchison (approved):
Thanks for fixing this.

PR #385, opened 2026-06-04
feat: add attachment support to gmail.createDraft tool

### LINKED ISSUE #359: Add Attachment Support to gmail.createDraft Tool
# Feature Request: Add Attachment Support to `gmail.createDraft` Tool

## Summary
Extend the `gmail.createDraft` tool to support attaching local files. This enables users to create drafts with one or more attachments by providing absolute file paths.

## Motivation
The current `gmail.createDraft` implementation only allows setting the recipient, subject, and body (text/HTML). In many workflows, users need to send files (reports, images, logs) along with their emails. Adding attachment support makes the tool much more versatile and aligns it with the capabilities of the Gmail API.

## Proposed Changes
1.  **Tool Schema Update**: Add an `attachments` array parameter to the `gmail.createDraft` tool. Each attachment object should include:
    - `filePath` (required): Absolute path to the local file.
    - `filename` (optional): Display name for the attachment.
    - `mimeType` (optional): MIME type of the file.
2.  **MIME Handling**: Enhance `MimeHelper` to construct `multipart/mixed` messages when attachments are present. Ensure `In-Reply-To` and `References` headers are preserved for threaded drafts.
3.  **File Operations**: Implement file reading and automatic MIME type inference based on file extensions in `GmailService`.
4.  **Validation**: Ensure absolute paths are used and handle file-not-found errors gracefully.

## Benefits
- Enables automated email workflows that involve file sharing.
- Provides a seamless way to reply to threads with attachments.
- Improves parity with the full Gmail web/mobile experience.

## Implementation Status
A draft implementation is available in **PR #354**, including comprehensive tests for:
- Basic attachment creation.
- Multiple attachments.
- MIME type inference.
- Threading support with attachments.
- Error handling for missing or relative paths.

mickey-mikey:
Supersedes #354 (auto-closed when the source fork was deleted). Closes #359.

Adds an optional `attachments` parameter to `gmail.createDraft`, accepting absolute local file paths with optional filename override and MIME type inference from extension.

## Changes since #354 — all gemini-code-assist review comments addressed

- **(high)** Total attachment size cap lowered 25MB → **18MB**: Gmail's 25MB limit applies to the whole MIME message and base64 inflates binary content ~33%, so raw bytes are capped with headroom for encoding overhead and headers.
- **(high)** `createDraft` now validates `to`/`cc`/`bcc`/`replyTo` via `emailArraySchema`, matching `send` (returns the same structured `Invalid email address format` error).
- **(medium)** Attachment `fs.stat` checks parallelised with `Promise.all`; sizes summed before any file content is read (fail-early preserved).
- **(medium, carried from #354's final revision)** stat-before-read validation and `replyTo` parity in `MimeHelper.createMimeMessageWithAttachments` were already incorporated.

## Tests

528/528 passing, including new coverage for createDraft email validation, attachment size enforcement, multiple attachments, MIME inference, and threaded-reply drafts with attachments.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/workspace-server/src/index.ts
+++ b/workspace-server/src/index.ts
@@ -92,6 +92,10 @@ const emailComposeSchema = {
     .union([z.string(), z.array(z.string())])
     .optional()
     .describe('BCC recipient email address(es).'),
+  replyTo: z
+    .string()
+    .optional()
+    .describe('The email address to which replies should be sent.'),
   isHtml: z
     .boolean()
     .optional()
@@ -1298,6 +1302,32 @@ System labels that can be modified:
           .optional()
           .describe(
             'The thread ID to create the draft as a reply to. When provided, the draft will be linked to the existing thread with appropriate reply headers.',
+          ),
+        attachments: z
+          .array(
+            z.object({
+              filePath: z
+                .string()
+                .describe(
+                  'Absolute local filesystem path to the file to attach (e.g., "/Users/name/downloads/report.pdf"). Use gmail.downloadAttachment first to save an email attachment locally before referencing it here.',
+                ),
+              filename: z
+                .string()
+                .optional()
+                .describe(
+                  'Display name for the attachment in the email. Defaults to the filename portion of filePath.',
+                ),
+              mimeType: z
+                .string()
+                .optional()
+                .describe(
+                  'MIME type of the attachment (e.g., "application/pdf"). Inferred from the file extension when omitted; falls back to "application/octet-stream".',
+                ),
+            }),
+          )
+          .optional()
+          .describe(
+            'Files to attach to the draft. Each entry must reference an absolute local path. Download attachments first with gmail.downloadAttachment if needed.',
           ),
       },
     },
--- a/workspace-server/src/index.ts
+++ b/workspace-server/src/index.ts
@@ -20,6 +20,7 @@ import { PeopleService } from './services/PeopleService';
 import { SlidesService } from './services/SlidesService';
 import { SheetsService } from './services/SheetsService';
 import { GMAIL_SEARCH_MAX_RESULTS } from './utils/constants';
+import { gmailAttachmentSchema } from './utils/validation';
 
 import { setLoggingEnabled, logToFile } from './utils/logger';
 import { applyToolNameNormalization } from './utils/tool-normalization';
@@ -1304,27 +1305,7 @@ System labels that can be modified:
             'The thread ID to create the draft as a reply to. When provided, the draft will be linked to the existing thread with appropriate reply headers.',
           ),
         attachments: z
-          .array(
-            z.object({
-              filePath: z
-                .string()
-                .describe(
-                  'Absolute local filesystem path to the file to attach (e.g., "/Users/name/downloads/report.pdf"). Use gmail.downloadAttachment first to save an email attachment locally before referencing it here.',
-                ),
-              filename: z
-                .string()
-                .optional()
-                .describe(
-                  'Display name for the attachment in the email. Defaults to the filename portion of filePath.',
-                ),
-              mimeType: z
-                .string()
-                .optional()
-                .describe(
-                  'MIME type of the attachment (e.g., "application/pdf"). Inferred from the file extension when omitted; falls back to "application/octet-stream".',
-                ),
-            }),
-          )
+          .array(gmailAttachmentSchema)
           .optional()
           .describe(
             'Files to attach to the draft. Each entry must reference an absolute local path. Download attachments first with gmail.downloadAttachment if needed.',
--- a/workspace-server/src/services/GmailService.ts
+++ b/workspace-server/src/services/GmailService.ts
@@ -25,8 +25,7 @@ const EXTENSION_MIME_MAP: Record<string, string> = {
   '.docx':
     'application/vnd.openxmlformats-officedocument.wordprocessingml.document',
   '.xls': 'application/vnd.ms-excel',
-  '.xlsx':
-    'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
+  '.xlsx': 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
   '.ppt': 'application/vnd.ms-powerpoint',
   '.pptx':
     'application/vnd.openxmlformats-officedocument.presentationml.presentation',
@@ -457,6 +456,44 @@ export class GmailService {
     }
   };
 
+  /**
+   * Validates recipient email addresses shared by send and createDraft.
+   * Returns a structured error response when validation fails, or null when
+   * all provided addresses are valid.
+   */
+  private validateEmailAddresses({
+    to,
+    cc,
+    bcc,
+    replyTo,
+  }: {
+    to: string | string[];
+    cc?: string | string[];
+    bcc?: string | string[];
+    replyTo?: string;
+  }) {
+    try {
+      emailArraySchema.parse(to);
+      if (cc) emailArraySchema.parse(cc);
+      if (bcc) emailArraySchema.parse(bcc);
+      if (replyTo) emailArraySchema.parse(replyTo);
+      return null;
+    } catch (error) {
+      return {
+        content: [
+          {
+            type: 'text' as const,
+            text: JSON.stringify({
+              error: 'Invalid email address format',
+              details:
+                error instanceof Error ? error.message : 'Validation failed',
+            }),
+          },
+        ],
+      };
+    }
+  }
+
   public send = async ({
     to,
     subject,
@@ -468,24 +505,14 @@ export class GmailService {
   }: SendEmailParams) => {
     try {
       // Validate email addresses
-      try {
-        emailArraySchema.parse(to);
-        if (cc) emailArraySchema.parse(cc);
-        if (bcc) emailArraySchema.parse(bcc);
-        if (replyTo) emailArraySchema.parse(replyTo);
-      } catch (error) {
-        return {
-          content: [
-            {
-              type: 'text' as const,
-              text: JSON.stringify({
-                error: 'Invalid email address format',
-                details:
-                  error instanceof Error ? error.message : 'Validation failed',
-              }),
-            },
-          ],
-        };
+      const validationError = this.validateEmailAddresses({
+        to,
+        cc,
+        bcc,
+        replyTo,
+      });
+      if (validationError) {
+        return validationError;
       }
 
       logToFile(`Sending email to: ${to}, subject: ${subject}`);
@@ -546,24 +573,14 @@ export class GmailService {
   }: CreateDraftParams) => {
     try {
       // Validate email addresses
-      try {
-        emailArraySchema.parse(to);
-        if (cc) emailArraySchema.parse(cc);
-        if (bcc) emailArraySchema.parse(bcc);
-        if (replyTo) emailArraySchema.parse(replyTo);
-      } catch (error) {
-        return {
-          content: [
-            {
-              type: 'text' as const,
-              text: JSON.stringify({
-                error: 'Invalid email address format',
-                details:
-                  error instanceof Error ? error.message : 'Validation failed',
-              }),
-            },
-          ],
-        };
+      const validationError = this.validateEmailAddresses({
+        to,
+        cc,
+        bcc,
+        replyTo,
+      });
+      if (validationError) {
+        return validationError;
       }
 
       logToFile(`Creating draft - to: ${to}, subject: ${subject}`);
@@ -619,19 +636,20 @@ export class GmailService {
               );
             }
 
+            let stats;
             try {
-              const stats = await fs.stat(att.filePath);
-              if (!stats.isFile()) {
-                throw new Error(
-                  `Attachment path is not a file: ${att.filePath}`,
-                );
-              }
-              return stats.size;
+              stats = await fs.stat(att.filePath);
             } catch (statError) {
               throw new Error(
                 `Could not access attachment file ${att.filePath}: ${statError instanceof Error ? statError.message : String(statError)}`,
               );
             }
+
+            if (!stats.isFile()) {
+              throw new Error(`Attachment path is not a file: ${att.filePath}`);
+            }
+
+            return stats.size;
           }),
         );
         const totalSize = attachmentSizes.reduce((sum, size) => sum + size, 0);
--- a/workspace-server/src/services/GmailService.ts
+++ b/workspace-server/src/services/GmailService.ts
@@ -18,6 +18,51 @@ import {
 import { gaxiosOptions } from '../utils/GaxiosConfig';
 import { emailArraySchema } from '../utils/validation';
 
+// Extension to MIME type map for common file types
+const EXTENSION_MIME_MAP: Record<string, string> = {
+  '.pdf': 'application/pdf',
+  '.doc': 'application/msword',
+  '.docx':
+    'application/vnd.openxmlformats-officedocument.wordprocessingml.document',
+  '.xls': 'application/vnd.ms-excel',
+  '.xlsx':
+    'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet',
+  '.ppt': 'application/vnd.ms-powerpoint',
+  '.pptx':
+    'application/vnd.openxmlformats-officedocument.presentationml.presentation',
+  '.txt': 'text/plain',
+  '.csv': 'text/csv',
+  '.html': 'text/html',
+  '.htm': 'text/html',
+  '.json': 'application/json',
+  '.xml': 'application/xml',
+  '.zip': 'application/zip',
+  '.png': 'image/png',
+  '.jpg': 'image/jpeg',
+  '.jpeg': 'image/jpeg',
+  '.gif': 'image/gif',
+  '.svg': 'image/svg+xml',
+  '.mp4': 'video/mp4',
+  '.mp3': 'audio/mpeg',
+};
+
+// Maximum total size for all attachments. Gmail's 25MB limit applies to the
+// entire MIME message, and base64 encoding inflates binary data by ~33%, so
+// cap raw attachment bytes at 18MB to leave room for encoding overhead and
+// message headers/body.
+const MAX_TOTAL_ATTACHMENT_SIZE_BYTES = 18 * 1024 * 1024;
+
+function getMimeTypeFromExtension(filePath: string): string {
+  const ext = path.extname(filePath).toLowerCase();
+  return EXTENSION_MIME_MAP[ext] ?? 'application/octet-stream';
+}
+
+type AttachmentInput = {
+  filePath: string;
+  filename?: string;
+  mimeType?: string;
+};
+
 // Type definitions for email parameters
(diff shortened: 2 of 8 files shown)


workspace-server/src/services/GmailService.ts
```diff
@@ -534,16 +607,79 @@ export class GmailService {
...
+        // Read each file from disk
+        const resolvedAttachments = await Promise.all(
+          attachments.map(async (att) => {
+            const content = await fs.readFile(att.filePath);
+            return {
+              filename: att.filename ?? path.basename(att.filePath),
+              content,
+              contentType:
+                att.mimeType ?? getMimeTypeFromExtension(att.filePath),
+            };
```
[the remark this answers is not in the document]
mickey-mikey: Addressed in 12a623d — attachment filenames (and contentType, which had the same exposure) are now sanitized in MimeHelper before header interpolation: CR/LF and control characters stripped (header injection), backslashes and double quotes escaped (parameter injection). Includes a regression test with an injection payload.

workspace-server/src/services/GmailService.ts
```diff
@@ -534,16 +607,79 @@ export class GmailService {
...
+                throw new Error(
+                  `Attachment path is not a file: ${att.filePath}`,
+                );
+              }
+              return stats.size;
+            } catch (statError) {
+              throw new Error(
+                `Could not access attachment file ${att.filePath}: ${statError instanceof Error ? statError.message : String(statError)}`,
+              );
+            }
```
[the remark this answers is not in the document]
mickey-mikey: Addressed in 12a623d — fs.stat() is wrapped on its own and the isFile() check now runs outside that try/catch, so the 'not a file' error surfaces directly without being re-wrapped.

workspace-server/src/services/GmailService.ts
```diff
@@ -489,10 +539,33 @@ export class GmailService {
...
+              type: 'text' as const,
+              text: JSON.stringify({
+                error: 'Invalid email address format',
+                details:
+                  error instanceof Error ? error.message : 'Validation failed',
+              }),
+            },
+          ],
+        };
+      }
```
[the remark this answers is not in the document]
mickey-mikey: Addressed in 12a623d — extracted a private validateEmailAddresses() helper; both send and createDraft call it.

workspace-server/src/services/GmailService.ts
```diff
@@ -424,6 +472,7 @@ export class GmailService {
         emailArraySchema.parse(to);
         if (cc) emailArraySchema.parse(cc);
         if (bcc) emailArraySchema.parse(bcc);
+        if (replyTo) emailArraySchema.parse(replyTo);
```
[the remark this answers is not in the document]
mickey-mikey: Addressed in 12a623d — createDraft now uses the shared validateEmailAddresses() helper.

workspace-server/src/services/GmailService.ts
```diff
@@ -534,16 +624,80 @@ export class GmailService {
...
+
+        // Read each file from disk
+        const resolvedAttachments = await Promise.all(
+          attachments.map(async (att) => {
+            const content = await fs.readFile(att.filePath);
+            return {
+              filename: att.filename ?? path.basename(att.filePath),
+              content,
+              contentType:
+                att.mimeType ?? getMimeTypeFromExtension(att.filePath),
```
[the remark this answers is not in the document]
mickey-mikey: Addressed in 0a1f7f3 — switched both fallbacks to ||, with a regression test covering empty-string filename and mimeType.

workspace-server/src/services/GmailService.ts
```diff
@@ -534,16 +624,82 @@ export class GmailService {
...
+        if (totalSize > MAX_TOTAL_ATTACHMENT_SIZE_BYTES) {
+          throw new Error(
+            `Total attachment size (${(totalSize / 1024 / 1024).toFixed(2)}MB) exceeds the maximum allowed limit of ${MAX_TOTAL_ATTACHMENT_SIZE_BYTES / 1024 / 1024}MB.`,
+          );
+        }
+
+        // Read each file from disk
+        const resolvedAttachments = await Promise.all(
+          attachments.map(async (att) => {
+            const content = await fs.readFile(att.filePath);
```
allenhutchison: `fs.readFile` here isn't wrapped, unlike `fs.stat` ~25 lines up which produces a descriptive `Could not access attachment file <path>: <reason>`. A file deleted or perm-changed between stat and read (or `EMFILE` from the concurrent `Promise.all` reads) will surface as a bare Node errno (e.g. `EACCES: permission denied, open '/abs/path'`) with no indication it came from attaching a file to a draft. Suggest wrapping it identically:

```ts
let content: Buffer;
try {
  content = await fs.readFile(att.filePath);
} catch (readError) {
  throw new Error(
    `Could not read attachment file ${att.filePath}: ${readError instanceof Error ? readError.message : String(readError)}`,
  );
}
```

workspace-server/src/services/GmailService.ts
```diff
@@ -410,33 +456,63 @@ export class GmailService {
...
+      if (bcc) emailArraySchema.parse(bcc);
+      if (replyTo) emailArraySchema.parse(replyTo);
+      return null;
+    } catch (error) {
+      return {
+        content: [
+          {
+            type: 'text' as const,
+            text: JSON.stringify({
+              error: 'Invalid email address format',
```
allenhutchison, line 502: This catch hardcodes `error: 'Invalid email address format'` for *any* throw, but the `try` wraps four `emailArraySchema.parse(...)` calls. If anything other than a Zod validation error escapes (e.g. a misconfigured schema or an internal fault), it gets mislabeled — a user with perfectly valid addresses would be told their format is invalid, with no path to resolution. Suggest narrowing to Zod errors and rethrowing the rest so real faults surface through the outer handler:

```ts
} catch (error) {
  if (!(error instanceof ZodError)) throw error;
  return { /* structured invalid-email response */ };
}
```

Minor related note: this early-return logs nothing, while the success path uses `logToFile` — a `logToFile` line before returning would make rejected attempts observable server-side.

workspace-server/src/services/GmailService.ts
```diff
@@ -534,16 +624,82 @@ export class GmailService {
...
-        cc: cc ? (Array.isArray(cc) ? cc.join(', ') : cc) : undefined,
-        bcc: bcc ? (Array.isArray(bcc) ? bcc.join(', ') : bcc) : undefined,
-        isHtml,
-        inReplyTo,
-        references,
-      });
+      let mimeMessage: string;
+
+      if (attachments && attachments.length > 0) {
+        // Validate all paths are absolute and check file sizes before reading anything
```
allenhutchison, line 644: Minor / defense-in-depth: there's a TOCTOU window here — sizes are summed from `fs.stat` results and checked against the 18MB cap, then the files are re-opened with `fs.readFile` in a separate pass. A file can grow, or be swapped for a symlink-to-device / FIFO, between the two passes, so the `isFile()` and size checks are advisory at read time (a FIFO swap could even hang `readFile`). Server-controlled local paths make this low-risk, but worth at least a comment acknowledging the window, or re-summing `content.length` after the reads to make the cap authoritative.

workspace-server/src/services/GmailService.ts
```diff
@@ -18,18 +18,64 @@ import {
...
+  '.png': 'image/png',
+  '.jpg': 'image/jpeg',
+  '.jpeg': 'image/jpeg',
+  '.gif': 'image/gif',
+  '.svg': 'image/svg+xml',
+  '.mp4': 'video/mp4',
+  '.mp3': 'audio/mpeg',
+};
+
+// Maximum total size for all attachments. Gmail's 25MB limit applies to the
```
allenhutchison: The rationale is correct but the margin is tighter than the wording implies: 18MB of raw bytes becomes ~24MB after base64 inflation, leaving only ~1MB of Gmail's 25MB envelope for headers and body — not the comfortable headroom "leave room for" suggests. Consider stating it concretely, e.g. "18MB of raw bytes becomes ~24MB after base64, leaving ~1MB headroom under Gmail's 25MB message limit for headers and body." Also worth making explicit that 18MB is the *pre-encoding* raw cap (the check runs against `stats.size`).

workspace-server/src/utils/MimeHelper.ts
```diff
@@ -146,11 +176,20 @@ export class MimeHelper {
...
+        replyTo,
+        inReplyTo,
+        references,
         isHtml,
       });
     }
 
     // Multipart message with attachments
+    if (inReplyTo) {
+      messageParts.push(`In-Reply-To: ${inReplyTo}`);
```
allenhutchison: `inReplyTo` and `references` are interpolated raw here, while the new `sanitizeHeaderValue` is only applied to attachment `contentType`/`filename`. These values come from a fetched thread's `Message-ID`/`References` headers, so they're effectively server-controlled and the practical risk is low — but since this PR's stated goal includes header-injection defense, running `sanitizeHeaderValue` over `inReplyTo`/`references` (and ideally all raw-interpolated header values) would make the protection consistent.

workspace-server/src/utils/MimeHelper.ts
```diff
@@ -92,6 +92,27 @@ export class MimeHelper {
     return encodedMessage;
   }
 
+  /**
+   * Strips characters that would allow MIME header injection (CR/LF and
+   * other control characters) from a header value.
```
allenhutchison: Tiny accuracy nit: the regex `/[\r\n\x00-\x1f\x7f]/g` also strips `\x7f` (DEL), which isn't a C0 control character, so "other control characters" understates it slightly. Consider "removes CR, LF, and all other C0 control characters plus DEL (0x7f)." (The explicit `\r\n` in the regex is redundant since they're inside `\x00-\x1f`, but harmless/defensive.)

workspace-server/src/index.ts
```diff
@@ -1299,6 +1303,32 @@ System labels that can be modified:
           .describe(
             'The thread ID to create the draft as a reply to. When provided, the draft will be linked to the existing thread with appropriate reply headers.',
           ),
+        attachments: z
```
allenhutchison, line 1307: This zod `attachments` object and the TS `AttachmentInput` type in `GmailService.ts` are two independent, hand-maintained definitions of the same shape with no link between them. They agree today, but nothing keeps them in sync — adding/renaming a field in one silently diverges from the other (the handler cast papers over it). Consider a single source of truth: `export const attachmentInputSchema = z.object({...})` from a shared module and `type AttachmentInput = z.infer<typeof attachmentInputSchema>`. As a bonus, the `filePath` absolute-path and non-empty `filename`/`mimeType` invariants could move into the schema (`.refine(path.isAbsolute, …)`, `.min(1)`) so they're enforced at the tool boundary with structured per-field errors instead of runtime `throw`s deep in `createDraft`.

allenhutchison (commented):
## Review summary

Thanks for this — it's a well-constructed PR. The security instincts are strong (size cap enforced via `fs.stat` *before* any `fs.readFile`, absolute-path + `isFile()` enforcement, header/filename sanitization with backslash-before-quote escaping), the comments explain *why* rather than *what*, and the test split between route-verification (`GmailService.test.ts`) and real MIME output (`MimeHelper.test.ts`) is the right structure. Nothing here is merge-blocking. The notes below are refinements — I've left inline comments on specific lines and grouped the test-coverage gaps and a couple of cross-cutting items here.

### Test coverage gaps

- **`fs.stat` failure path is untested.** The only attachment-failure test rejects `fs.readFile`, but `beforeEach` always resolves `fs.stat`, so the dedicated stat catch and its `Could not access attachment file …` message never run. Add a test that rejects `fs.stat` and asserts that message plus that the Gmail API is not called.
- **Multiple-attachment summation is never exercised.** Every attachment test passes a single-element array, and the 18MB cap test trips with one 30MB file — so the `attachmentSizes.reduce(...)` sum is untested. A bug using `attachmentSizes[0]` or `Math.max` would pass. Add a two-file test (e.g. 2×10MB) asserting the combined total is rejected, and ideally a two-file success asserting both files appear in the `attachments` array passed to `createMimeMessageWithAttachments`.
- **Backslash-escaping branch in `sanitizeHeaderFilename` is untested.** The injection test covers quotes + CR/LF but never a literal backslash, so the `\\` → `\\\\` replace (which must run *before* the quote escape) is uncovered. Add a filename containing a backslash and assert it appears as `\\\\` in the decoded `Content-Disposition`.
- **Invalid `replyTo` (and `cc`/`bcc`) validation is unasserted.** `createDraft` invalid-email coverage only tests an invalid `to`. Since routing `replyTo` through `validateEmailAddresses` is part of this PR's purpose, add a case with a malformed `replyTo` asserting `Invalid email address format`.

### Cross-cutting design note

- **`send` does not support attachments / `inReplyTo` / `references`, while `createDraft` does.** Intentional scope, just flagging the asymmetry — the building block (`createMimeMessageWithAttachments`) already exists and is only wired into `createDraft`.

### Strengths worth calling out

- Size validated before any file read — large files aren't loaded into memory only to be rejected.
- Absolute-path enforcement with no path joining, so traversal isn't applicable.
- `||` (not `??`) fallback for filename/mimeType is deliberate, correct, and well-commented.
- The header-injection test asserts exact decoded output for both CR/LF stripping and quote escaping — refactor-resilient and tied directly to the security goal.

_Review assisted by automated analysis agents; all findings verified against the diff._

workspace-server/src/services/GmailService.ts
mickey-mikey: Addressed in 0aabaff — fs.readFile is now wrapped identically to fs.stat, so read failures surface as `Could not read attachment file <path>: <reason>` instead of a bare errno. The existing ENOENT test now also asserts the wrapper prefix.

workspace-server/src/services/GmailService.ts
mickey-mikey: Addressed in 0aabaff — the catch now narrows to ZodError and rethrows anything else, so internal faults surface through the caller's generic handler instead of being mislabeled as invalid email format. Also added the logToFile line on the rejection path so rejected attempts are observable server-side.

workspace-server/src/services/GmailService.ts
mickey-mikey: Addressed in 0aabaff — went with the authoritative option: after the reads, the total is re-summed from `content.length` (the bytes actually read) and checked against the cap again, with a comment explaining that the stat pass is a pre-flight guard and the post-read check closes the TOCTOU window. Covered by a test where stat reports 1KB but readFile returns an oversized buffer.

workspace-server/src/services/GmailService.ts
mickey-mikey: Addressed in 0aabaff — reworded to state it concretely: the cap is on raw pre-encoding bytes (checked against stats.size, and now also against the bytes read), 18MB becomes ~24MB after base64, leaving ~1MB of headroom under Gmail's 25MB message limit for headers and body.

workspace-server/src/utils/MimeHelper.ts
mickey-mikey: Addressed in 0aabaff — `In-Reply-To` and `References` are now run through `sanitizeHeaderValue` in both `createMimeMessage` and `createMimeMessageWithAttachments`, with injection tests for both paths asserting CR/LF cannot start a new header line.

workspace-server/src/utils/MimeHelper.ts
mickey-mikey: Addressed in 0aabaff — comment now reads "CR, LF, and all other C0 control characters, plus DEL (0x7f)" and notes the explicit \r\n is redundant with \x00-\x1f but kept for clarity. (Also dropped the `no-control-regex` eslint-disable on that line — it was reported as an unused directive.)

workspace-server/src/index.ts
mickey-mikey: Addressed in 0aabaff — added a shared `gmailAttachmentSchema` in `utils/validation.ts` as the single source of truth: `index.ts` uses it directly in the tool schema (descriptions included) and `GmailService` derives `AttachmentInput` via `z.infer`, so the two can no longer drift. The absolute-path invariant moved into the schema via `.refine(path.isAbsolute)`, enforced at the tool boundary with a structured error (the runtime check in `createDraft` stays as defense in depth for direct callers).

One deliberate deviation: no `.min(1)` on `filename`/`mimeType` — empty strings are accepted at the boundary because the service intentionally falls back to path-derived defaults for them (the `||` behavior from the earlier review round, with its regression test). The schema doc comment notes this.

mickey-mikey:
Thanks for the thorough review @allenhutchison — all inline comments and the test-coverage gaps from the summary are addressed in 0aabaff (replies on each thread).

On the coverage gaps specifically:

- **fs.stat failure path** — added a test rejecting `fs.stat` with EACCES, asserting the `Could not access attachment file …` message and that neither `fs.readFile` nor the Gmail API is called.
- **Multiple-attachment summation** — added a 2×10MB rejection test (each file under the cap, only the sum exceeds it, asserting the summed `20.00MB` in the message) and a two-file success test asserting both files appear in order in the `attachments` array passed to `createMimeMessageWithAttachments`.
- **Backslash-escaping branch** — added a MimeHelper test with a filename containing a literal backslash and a quote, asserting the exact decoded `Content-Disposition` (`\` before `\"`, so escape order is pinned).
- **Invalid `replyTo`/`cc`/`bcc`** — added a malformed-`replyTo` case and a malformed `cc`+`bcc` case, both asserting `Invalid email address format` with no MIME construction.

On the cross-cutting note about `send` not supporting attachments/`inReplyTo`/`references`: intentional scope for this PR — the building block is in place and wiring it into `send` would be a natural follow-up.

Full suite: 542 tests passing, lint and typecheck clean.

allenhutchison:
## Follow-up review — all comments addressed ✅

Thanks for the thorough turnaround on commit `0aabaff`. I re-verified each item against the actual code (not just the replies), and all seven are genuinely resolved — several going a step beyond what was suggested:

- **`fs.readFile` wrapping** — read failures now surface as `Could not read attachment file <path>: <reason>`, matching the `fs.stat` treatment.
- **Email-validation catch** — narrows to `ZodError` and rethrows internal faults, plus the added `logToFile` on the rejection path makes rejected attempts observable.
- **Header sanitization** — `In-Reply-To`/`References` now run through `sanitizeHeaderValue` in both MIME paths, with injection tests for each.
- **TOCTOU** — went with the authoritative fix: re-summing `content.length` after the reads via `assertWithinAttachmentSizeLimit` closes the window rather than just documenting it.
- **Shared schema** — `gmailAttachmentSchema` as the single source of truth (`index.ts` consumes it, `AttachmentInput` derived via `z.infer`), and the `.refine(path.isAbsolute)` now enforces the absolute-path invariant at the boundary — nice touch.
- **Comment accuracy** — both the 18MB-margin and DEL (0x7f) wordings are now precise.

Build health on my end: all 28 test suites pass (**542 tests**, up from 528 — good regression coverage on the new branches), and `npm run lint` is clean.

Nothing outstanding from my side. 🚀

allenhutchison (approved):

